PR #90, opened 2015-01-24
Migrated Markdown to AsciiDoc

bodiam:
This is the initial migration from Markdown to AsciiDoc. All documentation has been automatically converted using https://github.com/bodiam/markdown-to-asciidoc, and is manually verified. 

All links seem to be working, including the external links which came from the `references.markdown` file.


erdi:
Thanks for the contribution.

PR #91, opened 2015-02-05
Fix for GEB-358 at checking for frame page switching

vijaybolleypally:
Fixed issue in such a way that:
1.Frame Page with at block: Page context will change to given frame page based on its at verification result
2.Frame Page without at block: Simple page context changes to given frame page(No UndefinedAtCheckerException thrown to user)


erdi:
Thanks for the contribution. I will hopefully pull it tonight.

erdi:
This has been merged, thanks.

PR #93, opened 2015-02-16
Suport for JUnit4 - screenshot on Error

ntotomanov-taulia:
@erdi  Kindly asking for review.
I've read that Geb is not currently allowing creating a report on test failure.
I've needed such support for my project and thought that maybe I can share it with you guys so it can be included in the Geb release.

Most of our tests were having a (Logout) in the @After methods so in case of error in smoke test - the logout was clearing the problematic page and screenshot was taken on the Logout from GebReporting class.

Added two new annotations to the JUnit4 rule chain:
@GebRule - support Field and Method annotation. Will be executed Before all @After annotated methods
@GebAfter -support Method annotation. Will be executed Before all @After annotated methods

GebSpringJUnit4ClassRunner - extends SpringJUnit4ClassRunner with support of the above annotation and rule chains
GebBlockJUnit4ClassRunner - extends BlockJUnit4ClassRunner with support of the above annotation and rule chains

Example Usage:
```
@RunWith(GebBlockJUnit4ClassRunner.class)
abstract class PlainGebTestCase extends GebTest {
  @Rule
  public TestName reportingTestName = new TestName()

  @GebRule
  public ScreenShotOnError screenShotOnError = new ScreenShotOnError(getClass(), reportingTestName)
}
```

### DIFF
--- a/module/geb-junit4/src/main/groovy/geb/junit4/ScreenShotOnError.groovy
+++ b/module/geb-junit4/src/main/groovy/geb/junit4/ScreenShotOnError.groovy
@@ -1,0 +1,38 @@
+package geb.junit4
+
+import geb.report.ReporterSupport
+import groovy.util.logging.Log4j
+import org.junit.rules.TestName
+import org.junit.rules.TestWatcher
+import org.junit.runner.Description
+
+@Log4j
+public class ScreenShotOnError extends TestWatcher{
+
+  @Delegate public GebReportingTest gebReportingTest = new GebReportingTest()
+
+  private TestName reportingTestName
+
+  private Class clazz
+
+  public ScreenShotOnError(Class clazz, TestName gebReportingTestName){
+    this.clazz = clazz
+    this.reportingTestName = gebReportingTestName
+  }
+
+  @Override
+  protected void failed(Throwable e, Description description) {
+    generateReport()
+  }
+
+  @Override
+  protected void starting(Description description) {
+    gebReportingTest.reportGroup clazz
+    gebReportingTest.incrementTestCounterValue()
+  }
+
+  public void generateReport(){
+    browser.report(ReporterSupport.toTestReportLabel(gebReportingTest.getTestCounterValue(), gebReportingTest.instanceTestCounter++,  reportingTestName.methodName, 'error'))
+  }
+
+}
--- a/module/geb-junit4/src/main/groovy/geb/junit4/runners/GebAfter.java
+++ b/module/geb-junit4/src/main/groovy/geb/junit4/runners/GebAfter.java
@@ -1,0 +1,11 @@
+package geb.junit4.runners;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD})
+public @interface GebAfter {
+}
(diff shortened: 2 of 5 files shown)


menonvarun:
Hi @ntotomanov-taulia,

I suppose this related to geb issue "GEB-92"
The fix for the issue can be as simple as following:

    @Rule
    public TestName gebReportingTestTestName = new TestName() {
		@Override
		protected void failed(Throwable e, Description description) {
			report "end"
		}

		@Override
		protected void succeeded(Description description) {
			if (!getConfig().reportOnTestFailureOnly) {
				report "end"
			}
		}
	}

@vijaybolleypally is working on the same fix for all unsupported plugins and will be ready in a day or so.

Please let me know if I am missing something here.

ntotomanov-taulia:
@menonvarun Thank you for the reply.

I've tried @Rule but the @Rule annotated fields/methods are below in JUnit RuleChain: 

```
 protected Statement methodBlock(FrameworkMethod method) {
        Object test;
        try {
            test = new ReflectiveCallable() {
                @Override
                protected Object runReflectiveCall() throws Throwable {
                    return createTest();
                }
            }.run();
        } catch (Throwable e) {
            return new Fail(e);
        }

        Statement statement = methodInvoker(method, test);
        statement = possiblyExpectingExceptions(method, test, statement);
        statement = withPotentialTimeout(method, test, statement);
        statement = withBefores(method, test, statement);
        statement = withAfters(method, test, statement);
        statement = withRules(method, test, statement);
        return statement;
    }
```

Your approach would work in scenarios when @After methods do not mess up the scenario but very often you have to move the LogOut in an @After method so to be sure that you do not leave sessions behind in case test failures.
When that happens - the @Rule will be executed after all @After annotated methods and we miss the HTML/screenshot capture.

I am proposing two new annotations which are up added UP in the RuleChain so all @GebRule and @GebAfter annotated fields/methods will be executed before @Afters (before the logout) and the screenshot/html capture will be taken on the problematic place. 

Not sure weather I was able to explain the scenario properly...

menonvarun:
This seems to be a new feature change rather than just targeting the existing issue. I will go in-depth to your solution tomorrow my time and will respond to you.

ntotomanov-taulia:
@menonvarun thank you

erdi:
Thanks for the contribution @ntotomanov-taulia but I don't think that we will be pulling this one in, at least in its current form.

First of all we don't want to depend on anything Spring related, so `GebSpringJUnit4ClassRunner` is out of the picture straight away. Secondly I think that custom JUnit runners do not belong to Geb's codebase.

I see that there are two things here - taking snapshots on error and being able to take them before any `@After` annotated methods. We could pull your implementation of `TestWatcher` for the first part but the second part is not really related to Geb, it's more a generic JUnit code and belongs elsewhere. There is also a solution for which you wouldn't need any of that custom JUnit code - just move your logout code to a rule and wrap it around the screenshot-on-error-taking `TestWatcher` using a [`RuleChain`](http://junit.org/javadoc/latest/org/junit/rules/RuleChain.html).

The `TestWatcher` implementation would still need some polishing (you will want to pass it a `GebTest` instead of instantiating one, add some test coverage for it and mention the usage in the Book of Geb) but it's definitely something we would be interested in. Given that apparently @vijaybolleypally is working on that as well you guys probably should synchronise so that you don't double the effort.

@menonvarun, @vijaybolleypally may I ask you guys to drop an email to the user or dev list before you pick up any contributions in the future please? It will allow me to give you guys some pointers and avoid situations where multiple people work on the same effort. It's always a shame to waste anyones effort.

menonvarun:
@erdi Thanks for your response, we will make sure to send a mail to the group before starting on any task.

erdi:
Closing due to no response from the requestor about suggested changes.

Please feel free to open a new request if you would still like to contribute a revised version of a `TestWatcher` implementation that would take screenshots on failures only.

PR #95, opened 2015-03-02
Improved error message for page method for list of page class

vijaybolleypally:
GEB-346:Improve error reporting when no at checkers pass if using multiple candidates for page switching
Added Code Fix and unit tests


module/geb-core/src/main/groovy/geb/error/UnexpectedPageException.java
```diff
@@ -52,4 +52,8 @@ public UnexpectedPageException(Class<? extends Page> actualPage, Class<? extends
     public UnexpectedPageException(Class<? extends Page> actualPage, Page[] potentials) {
         super(String.format("An unexpected page %s was encountered when trying to find page match (given potentials: %s)", actualPage.getName(), DefaultGroovyMethods.toString(potentials)));
     }
+
+	public UnexpectedPageException(List<Class<? extends Page>> potentials, List<Throwable> exceptionDetails) {
```
erdi: Indentation seems to be off here.

module/geb-core/src/main/groovy/geb/Page.groovy
```diff
@@ -206,6 +206,26 @@ class Page implements Navigable, PageContentContainer, Initializable {
...
     /**
+     * Executes this page's "at checker" and returns caused exception.
+     *
+     * @param honourGlobalAtCheckWaiting
+     * @return exception i.e thrown while verifying at checker for current page (returns null if at verification is success)
+     */
+    Throwable verifyAtAndReturnCaughtException(boolean honourGlobalAtCheckWaiting = true) {
+        Throwable caughtException = null
+        try {
+            verifyThisPageAtOnly(honourGlobalAtCheckWaiting)
```
erdi: This will unfortunately not work with implicit assertions disabled and we still want to support that use case. You can simulate disabling implicit assertions of at block in tests [this way](https://github.com/geb/geb/blob/master/module/geb-core/src/test/groovy/geb/PageOrientedSpec.groovy#L476).

I would suggest returning a value type from this method that would take value returned from `verifyThisPageAtOnly()` and the exception caught as constructor parameters and would have a single method (`passed`?) that would return true if there was no exception thrown and verification has passed. We could call that type `AtVerificationResult`. The implementation of `verifyAtSafely()` should also use this method to avoid duplication of logic.

module/geb-core/src/test/groovy/geb/UnexpectedPagesSpec.groovy
```diff
@@ -210,6 +210,20 @@ class UnexpectedPagesSpec extends GebSpecWithCallbackServer {
         "foo"          | 'foo'
         UnexpectedPage | "class geb.UnexpectedPage"
     }
+
+    void "ensure that an exception message with all page wise error details is thrown when no match is found in given list of pages"() {
```
erdi: We should move this test to `PageOrientedSpec`. This spec is about verification of behaviour around [unexpected pages](http://www.gebish.org/manual/current/pages.html#unexpected_pages).

module/geb-core/src/main/groovy/geb/error/UnexpectedPageException.java
```diff
@@ -52,4 +52,8 @@ public UnexpectedPageException(Class<? extends Page> actualPage, Class<? extends
...
+		super(String.format("Unable to find page match (given potentials: %s) failed with (respective errors :\n %s)", DefaultGroovyMethods.toString(potentials), DefaultGroovyMethods.toString(exceptionDetails)));
```
erdi: We can definitely do better than this when it comes to reporting. At the moment information is in multiple places - you have to mentally connect the list of pages with the list of failures because. We should print a page name and the result of it's evaluation (which doesn't necessarily have to be an exception, see my comment on `verifyAtAndReturnCaughtException()`)  together. We should show exception messages and first line of their stacktrace for `AssertionError`s and `RequiredPageContentNotPresent` exceptions and message of the ImplicitWaitTimeoutException and message of it's cause together with the first line of it's stacktrace.

erdi:
This is a good start but it would be cool if you could address my comments.

erdi:
There are still some issues with this. Do you want me to provide more feedback or are you happy for me to pull it and rework it a bit myself?

menonvarun:
Few things have been missed. Let us try to fix it from our side.

erdi:
Sure, take your time.

module/geb-core/src/main/groovy/geb/AtVerificationResult.groovy
```diff
@@ -0,0 +1,71 @@
...
+        } else if (errorThrown instanceof ImplicitWaitTimeoutException) {
+            ImplicitWaitTimeoutException timeoutException = (ImplicitWaitTimeoutException) errorThrown
+            message += "Timeout exception message: " + timeoutException.getMessage()
+            message += ": Cause message : " + timeoutException.getCause().getMessage()
+            if(timeoutException.getCause().getStackTrace().length > 0)
+                message += ": Stacktrace : " + timeoutException.getCause().getStackTrace()[0].toString()
+        } else {
+            message = DefaultGroovyMethods.toString(errorThrown)
+        }
+        message
```
menonvarun: Segregating the messages with distinguished string ex. "Exception message", "Cause message" etc.

module/geb-core/src/main/groovy/geb/AtVerificationResult.groovy
```diff
@@ -0,0 +1,71 @@
...
+            message += "Exception message: " + errorThrown.getMessage()
+            if(errorThrown.getStackTrace().length > 0)
+                message += ": Stacktrace : " + errorThrown.getStackTrace()[0].toString()
+
+        } else if (errorThrown instanceof ImplicitWaitTimeoutException) {
+            ImplicitWaitTimeoutException timeoutException = (ImplicitWaitTimeoutException) errorThrown
+            message += "Timeout exception message: " + timeoutException.getMessage()
+            message += ": Cause message : " + timeoutException.getCause().getMessage()
+            if(timeoutException.getCause().getStackTrace().length > 0)
+                message += ": Stacktrace : " + timeoutException.getCause().getStackTrace()[0].toString()
```
menonvarun: Segregating the messages with distinguished string ex. "Exception message", "Cause message" , "Stacktrace" etc.

menonvarun:
@erdi Worked on the pending issues based on your earlier comments. Please take a look at getErrorMessage method under the AtVerificationResult class where I had implemented the logic for appending exception message along with the cause and first line of StackTrace(if any) as I am not sure about how you want the exception to be presented to the user. Please let us know if there is anything more that needs to be worked upon.

erdi:
Thanks @menonvarun, I will try to have a look at it over the weekend and report back.

module/geb-core/src/main/groovy/geb/Page.groovy
```diff
@@ -194,15 +194,27 @@ class Page implements Navigable, PageContentContainer, Initializable {
...
+    /**
+     * Executes this page's "at checker" and sends result of execution to AtVerificationResult class.
+     *
+     * @param honourGlobalAtCheckWaiting
+     * @return instance of AtVerificationResult class which holds current page at result
+     */
+    AtVerificationResult verifyAtAndStoreCaughtException(boolean honourGlobalAtCheckWaiting = true) {
+        Throwable caughtException = null
         try {
             verifyThisPageAtOnly(honourGlobalAtCheckWaiting)
```
erdi: This still does not take the value returned from `verifyThisPageOnly()` into consideration. See my comments on the now outdated diffs that we should handle implicit assertions being disabled and what should happen then.

module/geb-core/src/main/groovy/geb/AtVerificationResult.groovy
```diff
@@ -0,0 +1,74 @@
...
+import geb.waiting.ImplicitWaitTimeoutException
+import org.codehaus.groovy.runtime.DefaultGroovyMethods
+
+/**
+ * Created by vijayb on 05-03-2015.
+ *
+ * Class to hold page data and its at verification result
+ */
+class AtVerificationResult {
+    private Page page
```
erdi: This field does not belong here. We should map pages to their verification result in the code that runs the checks for every elements in a list of page candidates and then pass this map to `UnexpectedPageException`. This class should merely hold a result of at verification, no need for it to know for which page that result is.

module/geb-core/src/main/groovy/geb/AtVerificationResult.groovy
```diff
@@ -0,0 +1,74 @@
...
+    /**
+     * Returns name of page class and cause for its at check failure
+     *
+     * @return
+     */
+    String toString() {
+        String.format("\n %s : %s", page.getClass().getName(), getErrorMessage())
+    }
+
+    private String getErrorMessage() {
```
erdi: This logic is correct but does not belong here - it should be moved to `UnexpectedPageException`. Presentation of the exception is not a concern of this class.

module/geb-core/src/main/groovy/geb/AtVerificationResult.groovy
```diff
@@ -0,0 +1,74 @@
...
+        this.page = page
+        this.errorThrown = errorThrown
+    }
+
+    /**
+     * Returns true if errorThrown is null(throwable object type)
+     *
+     * @return boolean value representing  at check verification result
+     */
+    boolean isPassed() {
```
erdi: Let's rename this method to `toBoolean()` - this class will then benefit from supporting groovy truth and it's instances will be able to be used in boolean expressions.

erdi:
I've added some comments and will let you apply them. Please be aware that I will still probably apply some more polishing after that.

menonvarun:
@erdi Made changes to the AtVerificationResult to support Implicit Assertions and changed the method name as per your suggestion. Still there are few things that is not resolved.
After adding method toBoolean inside AtVerifcationResult class still its object is not able to take the benefit of groovy truth and I am forced to call the "toBoolean" method.
Also still Page object is still stored inside the AtVerificationResultClass and the error message is still evaluated inside it. The message cannot be evaluated inside the UnexpectedPageException since the exception message can only be stored by calling the super exception constructor.

erdi:
You could use a private static method in `UnexpectedPageException` to get around the problem of having to pass it to the super constructor.

Sorry, it turns out that is has to be [`asBoolean()`](https://github.com/geb/geb/blob/master/module/geb-core/src/main/groovy/geb/navigator/AbstractNavigator.groovy#L56) and not `toBoolean()`.

I will pick this up in whatever state it will be when I'm done with migrating Geb issues from Jira@Codehaus to GitHub issues.

menonvarun:
@erdi Thanks for the response. I had made changes as per your response. 
This change turned out to be much more complex then what we predicted. Thanks for your effort and time in giving us your valuable feedback.

erdi:
No worries and thanks for bearing with me even when you're asked to change your contribution. Much appreciated!

As I said, it will be a while before I get around to merging it, but will definitely make into the next version. :)

I'm also happy to give pointers in the future to issues that are less complex. Just give me a shout on the dev list.

menonvarun:
No Issues. Take your time. 
According to me there is no harm in working on feedback because it helps in maintaining the code quality.

erdi:
Yes, code quality and consistency is the end goal here, indeed.

erdi:
Thanks for the contibutioon @vijaybolleypally, @menonvarun.

Sorry that it took so long to merge, I wanted to finish the migration of the Book of Geb to Asciidoctor before picking up any other work and now that it's done I'm catching up on PRs.

PR #96, opened 2015-03-02
Code fix for clear browser after each test in TestNG

vijaybolleypally:
Added code for clearing browser after each test and updated unit tests respectively


erdi:
Thanks for the contribution!

PR #97, opened 2015-03-08
Change mailing list information in README.md.

PoohSunny:
mailing list information has been updated in [lists.html](https://github.com/geb/geb/blob/master/doc/site/src/ratpack/templates/includes/lists.html). But seems not to be updated in README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,4 +60,4 @@ To run the entire test suite, run…
     
 ## Development Mailing List
 
-If you want to do some work on Geb and want some help, you can join the `dev@geb.codehaus.org` mailing list via [Codehaus' Xircles](http://xircles.codehaus.org/projects/geb/lists "Codehaus: Geb: Lists").
+If you want to do some work on Geb and want some help, you can join the `geb-dev@googlegroups.com` mailing list via [Google Groups](https://groups.google.com/d/forum/geb-dev).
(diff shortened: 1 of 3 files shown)


erdi:
Thanks for the contribution Yotaro. Well spotted.

This is not the only place where the address of the mailing list is incorrect. Would you also mind fixing it in the [manual](https://github.com/geb/geb/blob/master/doc/manual/src/chapters/140-project.md#support--development)?

Unfortunately we are in the process of migrating the Book of Geb from markdown to asciidoctor which means that it also need fixing [here](https://github.com/geb/geb/blob/master/doc/new-manual/src/docs/asciidoc/140-project.adoc#support--development).

It would be also helpful if you added yourself as a contributor [in the markdown version](https://github.com/geb/geb/blob/master/doc/manual/src/chapters/140-project.md#contributors) and [in the asciidoctor version](https://github.com/geb/geb/blob/master/doc/new-manual/src/docs/asciidoc/140-project.adoc#contributors). This way I could then pull it with one click of a button. :)

PoohSunny:
I changed what you pointed out :) Please let me know if it needs some other changes.

PoohSunny:
I don't understand why test fails. The test is passed when I run it in my local and It seems not to have relationship with my change... Can you please advise me how I can fix this...?

erdi:
Don't worry about it, that test is flakey. It's not related to your change.

PoohSunny:
@erdi Thank you :) I understand.

erdi:
Thanks for the contribution!

PR #98, opened 2015-04-01
When an option cannot be selected, include the available values in the exception message

jochenberger:
Currently, if an option cannot be selected from a `select` element, you get a trace like
```
java.lang.IllegalArgumentException: couldn't select option with text or value: foo
	at geb.navigator.NonEmptyNavigator.setSelectValue(NonEmptyNavigator.groovy:738)
	at geb.navigator.NonEmptyNavigator.setInputValue(NonEmptyNavigator.groovy:690)
	at geb.navigator.NonEmptyNavigator.setInputValues_closure42(NonEmptyNavigator.groovy:684)
	at groovy.lang.Closure.call(Closure.java:423)
	at groovy.lang.Closure.call(Closure.java:439)
	at geb.navigator.NonEmptyNavigator.setInputValues(NonEmptyNavigator.groovy:683)
	at geb.navigator.NonEmptyNavigator.value(NonEmptyNavigator.groovy:441)
	at geb.content.TemplateDerivedPageContent.value(TemplateDerivedPageContent.groovy:27)
```
It would help if the select's available option values were included in the error message.

### DIFF
--- a/module/geb-core/src/main/groovy/geb/navigator/NonEmptyNavigator.groovy
+++ b/module/geb-core/src/main/groovy/geb/navigator/NonEmptyNavigator.groovy
@@ -749,7 +749,7 @@ class NonEmptyNavigator extends AbstractNavigator {
                 try {
                     select.selectByVisibleText(valueString)
                 } catch (NoSuchElementException e2) {
-                    throw new IllegalArgumentException("couldn't select option with text or value: $valueString")
+                    throw new IllegalArgumentException("couldn't select option with text or value: $valueString, available values: $valueStrings")
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


jochenberger:
Oh, this PR is nonsense, I misunderstood what the `valueStrings` variable contains.

jochenberger:
That should do it.

erdi:
Can you please also include a test for this change, @jochenberger? It would go into `SelectControlSpec`.

jochenberger:
Sure. Would it be sufficient to just add an assertion on the exception message to an existing feature or should I create a separate one?
Unfortunately I cannot run the specific test isolated on my machine. `gradle module:geb-core:test --tests geb.navigator.SelectControlSpec` fails with lots of classpath errors. I'll try to run the whole suite...

jochenberger:
No, I'm not able to get the tests to run and currently I don't have the time to keep on trying. But I think that they should pass with the change I made.

erdi:
If you still want to run it then you should use the wrapper: 

`./gradlew module:geb-core:test --tests geb.navigator.SelectControlSpec`

But your test looks like it should be ok so you can leave it as it is.

erdi:
Thanks for the contribution.

PR #99, opened 2015-05-27
set forcelocal as default option

matankatz:
set forcelocal as default, and the option to disable it.

### DIFF
--- a/integration/geb-gradle/src/main/groovy/geb/gradle/browserstack/BrowserStackTunnel.groovy
+++ b/integration/geb-gradle/src/main/groovy/geb/gradle/browserstack/BrowserStackTunnel.groovy
@@ -28,11 +28,13 @@ class BrowserStackTunnel extends ExternalTunnel {
 
     final String outputPrefix = 'browserstack-tunnel'
     final String tunnelReadyMessage = 'You can now access your local server(s) in our remote browser.'
+    final boolean isForceLocal
 
-    BrowserStackTunnel(Project project, Logger logger, BrowserStackAccount account, List<URL> applicationUrls) {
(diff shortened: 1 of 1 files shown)


matankatz:
Here is a reference to browserstack doc:
https://www.browserstack.com/local-testing#whitelist-ip

integration/geb-gradle/src/main/groovy/geb/gradle/browserstack/BrowserStackTunnel.groovy
```diff
@@ -28,11 +28,13 @@ class BrowserStackTunnel extends ExternalTunnel {
 
     final String outputPrefix = 'browserstack-tunnel'
     final String tunnelReadyMessage = 'You can now access your local server(s) in our remote browser.'
+    final boolean isForceLocal
 
-    BrowserStackTunnel(Project project, Logger logger, BrowserStackAccount account, List<URL> applicationUrls) {
+    BrowserStackTunnel(Project project, Logger logger, BrowserStackAccount account, List<URL> applicationUrls, boolean isForceLocal = true) {
```
erdi: Given that in the past it was not set this would be a breaking change. We should default to false here.

erdi:
We will also want to add a field to `BrowserStackExtension` so that this can be configured via the DSL in the Gradle plugin. See how it's done for `applicationUrls` for example.

integration/geb-gradle/src/main/groovy/geb/gradle/browserstack/BrowserStackTunnel.groovy
erdi: Actually let's not use any default value for this parameter and default it to false in `BrowserStackExtension` as per my comment on the top level of this PR.

erdi:
Superseeded by https://github.com/geb/geb/pull/100

PR #100, opened 2015-05-31
support forcelocal same as application

matankatz:
opening tunnel with browser stack requires sometimes to use the local network.
adding the ability to set the flag -forcelocal to the tunnel executable.

in browserstack task just need to use:
forcelocal true


erdi:
Thanks for the contibutioon @matankatz.

Sorry that it took so long to merge, I wanted to finish the migration of the Book of Geb to Asciidoctor before picking up any other work and now that it's done I'm catching up on PRs.

matankatz:
Sure, glad i could contribute.

Thanks,
Matan
בתאריך 29 ביונ' 2015 12:28,‏ "Marcin Erdmann" <notifications@github.com>
כתב:

PR #101, opened 2015-06-11
add a Factory that restarts the browser with every usage

dnoseda:
I have a test batery that runs in parallel several browsers, and sometimes to clear the cookies is not enough. I need to quit that instance of the browser and to create a new one.

With this PR you can configurate this with the variable cacheDriverPerThreadWithReset

### DIFF
--- a/module/geb-core/src/main/groovy/geb/Configuration.groovy
+++ b/module/geb-core/src/main/groovy/geb/Configuration.groovy
@@ -197,6 +197,18 @@ class Configuration {
     }
 
     /**
+     * The driver is to be cached, this setting controls whether or not the driver is cached per thread but reseting the browser
+     * (quiting and opening a new instance),
+     * or per
+     * <p>
+     * The value is the config entry {@code cacheDriverPerThread}, which defaults to {@code true}.
+     */
+    boolean isCacheDriverPerThreadWithReset() {
+        readValue('cacheDriverPerThreadWithReset', false)
+    }
+
+
+    /**
      * The driver is to be cached, this setting controls whether or not the driver is cached per thread,
      * or per
      * <p>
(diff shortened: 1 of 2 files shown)


erdi:
Cheers for your contribution, @dnoseda. Can you please explain why it's needed, given that there is a `static CachingDriverFactory.clearCacheAndQuitDriver()` which allows you to "reset" the cached driver on demand and you could use it in `cleanup()` or `cleanupSpec()` or even better in an overridden version `GebSpec.createBrowser()`? 

I'm inclined not to accept it based on that.

dnoseda:
I tried clearCacheAndQuitDriver in the tearDown but it throws:

org.openqa.selenium.remote.SessionNotFoundException: Session ID is null.
Using WebDriver after calling quit().

I also tried a new createBrowser() but it has a similar effect

On Sat, Jun 13, 2015 at 7:15 AM, Marcin Erdmann <notifications@github.com>
wrote:

> Cheers for your contribution, @dnoseda <https://github.com/dnoseda>. Can
> you please explain why it's needed, given that there is a static
> CachingDriverFactory.clearCacheAndQuitDriver() which allows you to
> "reset" the cached driver on demand and you could use it in cleanup() or
> cleanupSpec() or even better in an overridden version
> GebSpec.createBrowser()?
>
> I'm inclined not to accept it based on that.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/geb/geb/pull/101#issuecomment-111695615>.
>



-- 
*Damian Noseda*
Ssr Developer
San Luis
cel.: (+54) 266 4378175

erdi:
I just tried and got it working this way:

    class DriverRecreatingGebSpec extends GebSpec {
        Browser createBrowser() {
            CachingDriverFactory.clearCacheAndQuitDriver()
            super.createBrowser
        }
    }

PR #103, opened 2015-06-25
Adding optional cause message to Wait time-out exceptions

victorparmar:
Users would like to see more information when a Wait time-out exception occurs and providing the cause of the exception improves productivity.

The output of Maven/Gradle error reporting tool is especially improved.

### DIFF
--- a/module/geb-core/src/main/groovy/geb/Configuration.groovy
+++ b/module/geb-core/src/main/groovy/geb/Configuration.groovy
@@ -94,7 +94,7 @@ class Configuration {
         } else if (waitingParam instanceof Number && waitingParam > 0) {
             getWait(waitingParam.doubleValue())
         } else if (waitingParam instanceof Collection) {
-            if (waitingParam.size() == 2 || waitingParam.size() == 3) {
+            if (waitingParam.size() > 1 && waitingParam.size() < 4) {
                 def timeout = waitingParam[0]
                 def retryInterval = waitingParam[1]
                 def printCause = getDefaultWaitPrintCause()
--- a/module/geb-core/src/main/groovy/geb/Configuration.groovy
+++ b/module/geb-core/src/main/groovy/geb/Configuration.groovy
@@ -73,16 +73,17 @@ class Configuration {
         def preset = rawConfig.waiting.presets[name]
         def timeout = readValue(preset, 'timeout', getDefaultWaitTimeout())
         def retryInterval = readValue(preset, 'retryInterval', getDefaultWaitRetryInterval())
+        def printCause = readValue(preset, 'printCause', getDefaultWaitPrintCause())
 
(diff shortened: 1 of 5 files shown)


erdi:
Thanks for the contribution @victorparmar.

Can you please explain why this would be useful given that the cause throwable's stacktrace is printed as part of the stacktrace for `WaitTimeoutException` anyway? With your solution it is just being printed twice - once as part of the message for `WaitTimeoutException` and once as part of the `Caused by` clause of the `WaitTimeoutException`.

sclassen:
Maybe this is than related to maven.
Our build currently prints a test failure in a waitFor as follows:

    -------------------------------------------------------
     T E S T S
    -------------------------------------------------------
    Running xxx.yyy.zzz.uitest.testManagement.AddNewStructureLevelSpec
    Tests run: 9, Failures: 0, Errors: 1, Skipped: 8, Time elapsed: 15.455 sec <<< FAILURE! - in xxx.yyy.zzz.uitest.testManagement.AddNewStructureLevelSpec
    Locate login page(xxx.yyy.zzz.uitest.testManagement.AddNewStructureLevelSpec)  Time elapsed: 12.877 sec  <<< ERROR!
    geb.waiting.WaitTimeoutException: condition did not pass in 10.0 seconds (failed with exception)
            at xxx.yyy.zzz.uitest.testManagement.AddNewStructureLevelSpec.Locate login page_closure3(AddNewStructureLevelSpec.groovy:51)
            at xxx.yyy.zzz.uitest.testManagement.AddNewStructureLevelSpec.Locate login page_closure3(AddNewStructureLevelSpec.groovy)
    
    
    Results :
    
    Tests in error: 
      AddNewStructureLevelSpec.Locate login page:51->GebSpec.methodMissing:54 » WaitTimeout
    
    Tests run: 9, Failures: 0, Errors: 1, Skipped: 8

The underlying cause is neither visible in the log nor in the surfire report. To find out why the wait is failing I need to catch the WaitTimeoutException in the test code and println the cause

erdi:
I first checked with the version of surefire plugin that is being used in https://github.com/geb/geb-example-maven (2.9 which is very old) and I could see that the whole stacktrace together with cause was being printed in the report. Then I checked with 2.18.1 and the cause is indeed nowhere to be seen. This makes this contribution useful. I'm so lucky not to have had to use Maven for anything serious in my life. :)

erdi:
Thanks for the contribution @victorparmar.

I have changed and simplified it slightly: https://github.com/geb/geb/commit/49662b60af20e131bf126021d98343560866dc8a. The main change is that the config option is now called `waiting.includeCauseInMessage`.

victorparmar:
@erdi, @sclassen  sounds good, thank you for working on it :)

PR #104, opened 2015-08-04
Change outdated link in release notes

pierre-hilt:
Links are dead in release note. I change it, but I have not rebuild the project to check if it is working properly.


erdi:
Thanks for spotting this @pierre-hilt. We have a spec that detects dead links on the page and in the manual but there was a massive threading related bug in it which I discovered thanks to you pointing out these dead links. I will pull this soon.

PR #105, opened 2015-08-08
[fix] delegate to the correct waitFor

berardino:
While upgrading to the latest geb version I found
that a lot of tests were failing. I then noticed that
the applied timeout was actually the interval.


erdi:
Thanks for spotting this and the contribution, @berardino.

erdi:
This is a partial fix for geb/issues#395.

PR #106, opened 2015-08-14
fix some API links in the documentation

madmas:
Fix API links to MultipleSelect, TextInput, FileInput and RadioButtons


erdi:
Thanks for the contribution. Looks like I've been copy pasting carelessly.

PR #107, opened 2015-08-19
Add support for non-ASCII letters in test names

ahitrin:
Hi!

It would be very useful to have readable test names for non-English speakers. But, currently, these names are mutated into the pile of underscore symbols in reports. Following simple change in escaping regexp should help.


erdi:
Thanks for the improvements and updating the test.

PR #108, opened 2015-09-18
fixes broken 'via()' link in manual

david-w-millar:
fixes broken 'via()' link in manual


erdi:
Thanks for spotting it and the contribution @david-w-millar. I added [checking for links that are broken because of syntax issues to `LinkCrawlSpec`](https://github.com/geb/geb/blob/97c1b25593041d49be727daf66c5ae3d2a470bc0/doc/site/src/test/groovy/geb/LinkCrawlSpec.groovy#L66-L68) so that issues like the one you fixed don't occur in the future. I've also noticed that you're already on the list of contributors.

david-w-millar:
My pleasure!

david-w-millar:
My pleasure.

PR #109, opened 2015-10-02
fix issue #407

kgeis:
Only the last class in a test suite would leave reports because they were being cleaned before each test class.


erdi:
Thanks for the PR but I won't be pulling it because I decided to change your fix slightly to still cleanup before every test but to cleanup the right directory. I also added some tests around it because this bug was a regression introduced in geb/geb@628c1b9250c14141da5b758fc776c7bfa13c0dfb. My fix and tests are in geb/geb@29c20e1 if you are interested.

PR #110, opened 2015-10-22
open up property to be able to access it within specs.

detlef-brendle:
In order to be able to save files within a Specification it is necessary to open up these properties to subclasses.

Usecase:
- my Specification class has got a method named 'my test method'
if I do report() a file is written with 001-001<some name>

In order to save a file within my test method it would be helpful to save it with the same prefix that the files from report() calls.

### DIFF
--- a/module/geb-spock/src/main/groovy/geb/spock/GebReportingSpec.groovy
+++ b/module/geb-spock/src/main/groovy/geb/spock/GebReportingSpec.groovy
@@ -24,9 +24,9 @@ class GebReportingSpec extends GebSpec {
     // Ridiculous name to avoid name clashes
     @Rule
     TestName gebReportingSpecTestName
-    private int gebReportingPerTestCounter = 1
+    protected int gebReportingPerTestCounter = 1
     @Shared
-    private int gebReportingSpecTestCounter = 1
+    protected int gebReportingSpecTestCounter = 1
 
     def setupSpec() {
         reportGroup getClass()


erdi:
Hi @detlef-brendle,

Can you please explain why you want to save files with the same prefix as the reports are using in more detail? Are you saving some files that are not reports? I would like to understand your usecase better before opening up these fields because at the moment I see them as an implementation detail.

detlef-brendle:
The reason for having a file named like the current report step is, that I want to save the content of an I-Frame which source is a PDF document. I would like to save this PDF beside the test-report.

I geb 0.10.0 this was possible, but now it is not anymore.
Maybe you could also provide a method that returns the reportFile prefix, so all Spec's can get this information.

Thanks,
Detlef

erdi:
Hi @detlef-brendle,

Sorry for the radio silence on this from my side.

After giving it some more consideration I will not be pulling this PR. These counters are an implementation detail and I don't want to expose them - people will start to depend on them and it will prevent me from being able to change them without imposing breaking changes. 

It feels like what you're after is a custom reporter (see [here for how to configure it](http://www.gebish.org/manual/current/#reporter-configuration)) that will check if the iframe with PDF as source is available in the page and write the report. If you use `geb.report.ReporterSupport` as the base of your reporter and use its `getFile(File, String, String)` method  then you won't have to know the location of the report directory or the values of the  counter fields you want to open up. See `geb.report.PageSourceReporter` and `geb.report.ScreenshotReporter` for how `ReporterSupport.getFile(File, String, String)` is supposed to be used. 

Please let me know if you have any additional questions.

detlef-brendle:
Hi @erdi ,

the reporter approach is fine, but i think it still would be nice to have the reporting state information available:

- I call downloadContent from downloadSupport to get a page content. This just retrieves the page content that is in my case a xml response.
- this response I want to store to the reporting files with the current step name.

e.g. 
- 001-001-0_1-methodName-end.html
- 001-001-0_1-methodName-XMLResponse.xml

would be nice to have an reporting API where I can pass in an additional String which represents the file content plus the file extension. something like this:
- report('XMLResponse','xml', '<xml><foo >test</foo></xml>')
that would result in the file '001-001-0_1-methodName-XMLResponse.xml'

what do you think ?

Thanks,
Detlef

erdi:
@detlef-brendle,

Yes, we could implement it the following way:
* Add a `Browser.report(String label, Reporter reporter)` method
* Add a `FileContentsReporter` implementation of `Reporter` that extends `ReporterSupport` and takes extension and data (in form of `byte[]`) as constructor arguments
* Add a `Browser.report(String label, String fileExtension, byte[] data)` method which instantiates a new `FileContentsReporter` and passes it to the new `Browser.report(String label, Reporter reporter)` method. Also add overloads that take data in the form of a `String` and `InputStream`
* Add methods to `geb.spec.GebReportingSpec` and `geb.junt4.GebTest` and `geb.testng.GebTest` that delegate to the new `report()` methods on `Browser`

detlef-brendle:
Thanks for providing 'createReportLabel()' on spec.
this solves my issue.

Well done,
Detlef

PR #111, opened 2015-11-13
missing import SauceAccount and BrowserStackAccount classes

alincode:
missing some import.


erdi:
Thanks for the contribution!

PR #92, opened 2015-02-11
GEB-287:Support for static atCheckWait on Page classes

vijaybolleypally:
Added code fix for GEB-287, updated unit tests and documentation respectively

PR #113, opened 2015-11-25
reportOnTestFailureOnly implementation for Spock (issue #92)

leo-13:
reportOnTestFailureOnly option is now working for Spock integration, without changing the Spock code. Implemented via JUnit MethodRule interface. If approved can be later used for Geb-JUnit integration as well.

### DIFF
--- a/module/geb-spock/src/main/groovy/geb/spock/GebReportingSpec.groovy
+++ b/module/geb-spock/src/main/groovy/geb/spock/GebReportingSpec.groovy
@@ -40,7 +40,7 @@ class GebReportingSpec extends GebSpec {
     }
 
     def cleanup() {
-        if (failTracker.failedTests.contains(gebReportingSpecTestName.methodName)) {
+        if (failTracker.failed) {
             report "failure"
         } else if (!browser.config.reportOnTestFailureOnly) {
             report "end"
--- a/module/geb-spock/src/main/groovy/geb/spock/GebReportingSpec.groovy
+++ b/module/geb-spock/src/main/groovy/geb/spock/GebReportingSpec.groovy
@@ -11,14 +11,13 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
- *
  */
 package geb.spock
 
 import geb.report.ReporterSupport
-import spock.lang.*
 import org.junit.Rule
 import org.junit.rules.TestName
+import spock.lang.Shared
 
 class GebReportingSpec extends GebSpec {
 
--- a/module/geb-spock/src/main/groovy/geb/spock/GebReportingSpec.groovy
+++ b/module/geb-spock/src/main/groovy/geb/spock/GebReportingSpec.groovy
@@ -15,9 +15,9 @@
 package geb.spock
 
 import geb.report.ReporterSupport
+import spock.lang.*
 import org.junit.Rule
 import org.junit.rules.TestName
-import spock.lang.Shared
 
 class GebReportingSpec extends GebSpec {
 
--- a/module/geb-spock/src/main/groovy/geb/spock/GebReportingSpec.groovy
+++ b/module/geb-spock/src/main/groovy/geb/spock/GebReportingSpec.groovy
@@ -40,14 +40,16 @@ class GebReportingSpec extends GebSpec {
     }
 
     def cleanup() {
-        report "end"
+        if (failTracker.failedTests.contains(gebReportingSpecTestName.methodName)) {
+            report "failure"
+        } else if (!browser.config.reportOnTestFailureOnly) {
+            report "end"
+        }
         ++gebReportingSpecTestCounter
     }
 
     void report(String label = "") {
-        if (!browser.config.reportOnTestFailureOnly || failTracker.failedTests.contains(gebReportingSpecTestName.methodName)) {
-            browser.report(ReporterSupport.toTestReportLabel(gebReportingSpecTestCounter, gebReportingPerTestCounter++, gebReportingSpecTestName.methodName, label))
-        }
+        browser.report(ReporterSupport.toTestReportLabel(gebReportingSpecTestCounter, gebReportingPerTestCounter++, gebReportingSpecTestName.methodName, label))
     }
 
 }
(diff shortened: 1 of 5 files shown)


ahitrin:
Looks like failed test has no relation to code in this pull request. At least, when I run tests locally, all have passed perfectly. Maybe, it was a temporary Ratpack's issue.

leo-13:
From what I see those are really random test failures

erdi:
Hi @leo-13,

Sorry for the radio silence, I was on holiday.

I had a custom Spock extension in the making for this but your solution is so much simpler that I think we should go with it. I will put some inline comments in a moment about the changes that I would like you to make, I hope that it's ok. Given you used a rule it could also be applied to the JUnit integration as you've mentioned and it would be awesome if you could incorporate it this PR.

I will now be way more responsive so it will get merged as soon as you apply the requested improvements and add it to the JUnit integration (if you feel like it, of course).

module/geb-spock/src/main/groovy/geb/spock/GebReportingSpec.groovy
```diff
@@ -43,7 +45,9 @@ class GebReportingSpec extends GebSpec {
     }
 
     void report(String label = "") {
-        browser.report(ReporterSupport.toTestReportLabel(gebReportingSpecTestCounter, gebReportingPerTestCounter++, gebReportingSpecTestName.methodName, label))
+        if (!browser.config.reportOnTestFailureOnly || failTracker.failedTests.contains(gebReportingSpecTestName.methodName)) {
```
erdi: This logic should be moved to the `cleanup()` method and put around the `report "end"` line. Probably also `failure` should be used as the label if there was a failure and `end` if there was no failure and `reportOnTestFailureOnly` is not enabled.

This method provides general reporting capability in the current test reporting group and can be used by the user directly and hence should not be changed.

module/geb-spock/src/main/groovy/geb/spock/MethodExecutionRule.groovy
```diff
@@ -0,0 +1,42 @@
...
+package geb.spock
+
+import org.junit.rules.MethodRule
+import org.junit.runners.model.FrameworkMethod
+import org.junit.runners.model.Statement
+
+/**
+ * Created by leo-13 on 11/11/2015.
+ */
+class MethodExecutionRule implements MethodRule {
```
erdi: This class can be simplified. We could extend from `org.junit.rules.TestWatcher` and override `void failed(Throwable e, Description description)` to set a `boolean` instance field to true when a failure happens. We don't need to track the name of the test because there will be a new instance of the rule created for each test.

module/geb-spock/src/test/groovy/geb/spock/GebReportingSpecSpec.groovy
```diff
@@ -65,6 +66,32 @@ class GebReportingSpecSpec extends GebReportingSpec {
...
+
+    def "reportOnTestFailureOnly is enabled - passing test"() {
+        given:
+        config.reportOnTestFailureOnly = true
+    }
+
+    def "reportOnTestFailureOnly is enabled - failing test"() {
+        given:
+        config.reportOnTestFailureOnly = true
+        failTracker.failedTests.add(specificationContext.currentIteration.name)
```
erdi: Oh, nice, I didn't know about `specificationContext`. On the other hand we won't need it if you apply the proposed changes to `MethodExecutionRule`.

module/geb-spock/src/test/groovy/geb/spock/GebReportingSpecSpec.groovy
```diff
@@ -65,6 +66,32 @@ class GebReportingSpecSpec extends GebReportingSpec {
...
+    }
+
+    def "there should be no report for the passing test when reportOnTestFailureOnly is enabled"() {
+        expect:
+        !reportGroupDir.listFiles().any { it.name.contains("passing test") }
```
erdi: Can we please match against the whole name of the test and not only part of it in this and the next test?

leo-13:
Hi Marcin,
Thanks for your feedback. I'm going to apply all the changes shortly. We were thinking about making it work for JUnit as well, but I couldn't figure out where to put MethodExecutionRule in order to reuse it from different modules. Can you please advise on that?

module/geb-spock/src/main/groovy/geb/spock/MethodExecutionRule.groovy
leo-13: I tried to implement your suggestion like that:
        boolean failed = false

    @Override
    protected void failed(Throwable e, Description description) {
        failed = true
    }

But it doesn't work. Failed tests are just ignored.

erdi:
Oh, right, I forgot that we'll need to share the JUnit rule across multiple modules. Therefore I will add a new module for that class and and change JUnit and Spock to both depend on it myself - it will be easier this way.

module/geb-spock/src/main/groovy/geb/spock/MethodExecutionRule.groovy
erdi: Did you change the class to extend `org.junit.rules.TestWatcher` instead of `MethodRule`?

leo-13: Yes I did, but it doesn't work as expected for some reason.

leo-13:
Ok, I applied of changes that you proposed, including removing the test names list. It's now a boolean variable. Please take a look at the final version.

erdi:
Ok, I will have a look, merge and change what's needed. Thanks for the contribution.

We [mention all contributors by first and last name in the book of Geb](https://github.com/geb/geb/blob/master/doc/manual/src/docs/asciidoc/140-project.adoc#contributors) but your github profile doesn't have a name associated with it. What name should I put in?

leo-13:
It's Leo Fedorov. Thanks Marcin, it's a pleasure to contribute at least a little bit. I've been using Geb for almost two years now and I really like it.

erdi:
Nice to hear that!

erdi:
Thanks again, this has been now merged.

I've applied some minor changes to it after merging and also used it for supporting `reportOnTestFailureOnly` config option in JUnit 4 integration.

leo-13:
Nice!

PR #115, opened 2016-02-17
GEB-422: HashMap.Values class in Groovy versions before 2.4.0 doesn't have a 'minus' method - converting it to a List to work with Groovy versions before 2.4.0

craigatk:
Background info in https://github.com/geb/issues/issues/422

To fully test that this fix works with Groovy versions < 2.4.0 I had to change the Groovy version in geb-grails.gradle from "project.groovyVersion" (2.4.5) to "2.3.7" (2.3.7 is the Groovy version that ships with Grails 2.4.2 which is Grails version used in geb-grails.gradle). I thought about committing that change but I didn't know if it would affect building/publishing the Grails plugin so I left it out. If you'd prefer that I commit that change just let me know and I'll include it. Or if you know of a way to set the Groovy version for the tests only that might work as well.

### DIFF
--- a/geb.gradle
+++ b/geb.gradle
@@ -101,17 +101,6 @@ subprojects {
 
         sourceCompatibility = 1.7
 
-        tasks.withType(Test) {
-            reports.junitXml.destination = reporting.file("test-results/$name")
-            reports.html.destination = reporting.file("test-reports/$name")
-            binResultsDir = file("$buildDir/test-results-bin/$name")
-
-            ext.driver = project.properties["driver"] ?: "htmlunit"
-            systemProperty 'geb.dev.driver', driver
-
-            systemProperty 'geb.build.reportsDir', reporting.file("geb-reports/$name")
-        }
-
         codenarc.configFile = rootProject.file('gradle/codenarc/rulesets.groovy')
 
         project.tasks.withType(CodeNarc) { codenarcSourceSetTask ->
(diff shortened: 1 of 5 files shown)


craigatk:
Ahh good point that it's a Groovy compatibility issue and not a Grails one.

Adding a new Groovy compatibility subproject and corresponding test sounds like a fun challenge! I'll give it a try tomorrow but if I run into too many issues I'll probably defer back to you :)

erdi:
Awesome. As I said, I probably won't have time to look into it before Friday evening so you can take your time. Please let me know tomorrow how it went.

geb.gradle
```diff
@@ -153,6 +142,17 @@ subprojects {
         }
     }
 
+    tasks.withType(Test) {
```
craigatk: I needed just the test task configuration section of the groovyModules configuration applied to the new compatibility subproject (the groovyModules section does other things like set an explicit Groovy 2.4.5 dependency that I don't want for the compatibility subproject), so I moved the test task configuration out of the groovyModules block. My Gradle skills aren't the best, so if there is a better way to accomplish this just let me know.

craigatk:
@erdi I've got a first attempt at creating the compatibility subproject with a test to verify Groovy 2.3.7 compatibility. When you get a chance to look at it let me know if there's anything I should do differently. Thanks!

erdi:
This will need some small tweaks but I will take care of that. Thanks, it looks good. I will merge it by the end of the weekend.

craigatk:
Sounds good, thanks Marcin!

erdi:
Thanks again for finding and fixing the issue. I made some changes to the build side of things introduced by this PR. You can see what they were by taking a look at https://github.com/geb/geb/commit/80bb71101a3fb4dc43f9910d399b9b4e8b0fea59 and https://github.com/geb/geb/commit/3f4fd6193586e87f9f81abd92fc2131feda70d77 if you are interested. I will push a minor release containing this fix in the coming days.

craigatk:
Cool I hadn't seen that way to override the versions using the resolution strategy before, that's pretty slick. Thanks for working on a release with this fix!

PR #119, opened 2016-04-19
Content wrapper can be cast to a delegate type

topr:
Gives the ability of unwrapping a module to its declared type. This is an improvement which comes as a result of discussion under the issue https://github.com/geb/issues/issues/433

Surely not a final solution yet but leaves the decision to the user: strong module typing or better error messages.

### DIFF
--- a/doc/manual-snippets/src/test/groovy/modules/UnwrappingModulesSnippetSpec.groovy
+++ b/doc/manual-snippets/src/test/groovy/modules/UnwrappingModulesSnippetSpec.groovy
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package modules
+
+import fixture.Browser
+import org.codehaus.groovy.runtime.typehandling.GroovyCastException
+
+class UnwrappingModulesSnippetSpec extends FormContentSpec {
+
+    @SuppressWarnings('UnusedVariable')
+    def "assignment of a module to a variable of its declared type fails"() {
+        when:
+        //tag::module_variable_fail[]
+        Browser.drive {
+            to ModulePage
+            FormModule foo = form   // GroovyCastException is thrown
+        }
+        //end::module_variable_fail[]
+        then:
+        thrown(GroovyCastException)
+    }
+
+    def "method invocation with an argument of module's declared type fails"() {
+        when:
+        //tag::module_argument_fail[]
+        Browser.drive {
+            to ModulePage
+            submitForm(form)   // MissingMethodException is thrown
+        }
+        //end::module_argument_fail[]
+        then:
+        thrown(MissingMethodException)
+    }
+
+    def "unwrapped module may be used with its declared type"() {
+        when:
+        //tag::module_cast[]
+        Browser.drive {
+            to ModulePage
+            FormModule foo = form as FormModule
+            submitForm(foo)
+        }
+        //end::module_cast[]
+        then:
+        noExceptionThrown()
+    }
+
+    //tag::module_argument_fail[]
+
+        void submitForm(FormModule formModule) {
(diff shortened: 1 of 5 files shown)


erdi:
@topr Don't worry about the failure, it's not related to your change. We test against a high number of WebDriver implementations and these test are sometimes flakey due to a number of reasons like loosing connectivity between the test machine and BrowserStack/SauceLabs, browser crashing, etc. And of course this happens most often when testing against IE...

The code and coverage looks good to me. Can you please add yourself to [the contributors](https://github.com/geb/geb/blob/master/doc/manual/src/docs/asciidoc/140-project.adoc#contributors) (if you wondered then you can add commits to a PR by simply pushing to the PR branch)? 

Also, if you can spare a bit more time then it would be awesome if you could add a short entry to the manual about the feature you're adding. Probably the best place for it would be in a new section at the end of the "Modules" chapter titled "Unwrapping modules returned from content DSL". Some short explanation and an example showing where the feature is useful is what I'm after. Please bear in mind that all snippets in the manual are tested and are coming from `:doc:manual-snippets` module. Feel free to let me know if you'd rather I did it.

topr:
After a few fixes all was green on local env with `gradle check`. I've noticed that the failing one is not executed by default and it's an IE variation. Well, IE so I thought most likely it's a flaky one :bowtie: 

I don't mind adding the corresponding doc, I'll do it soon. Thanks!

topr:
@erdi All right, it's done. Please have a look and let me know if it's ok.

erdi:
I tweaked the docs a bit, but still, this is a quality contribution. Thank you very much.

topr:
Cool, thanks!
I see the tweaks, I was wondering if reusing snippet classes from other example is a good idea actually.
Anyway, I'm glad you liked it and the PR is merged.

Is there any binary with this change available yet? Or when it may be expected to be (snapshot os ok for the time being)?

erdi:
Yeah, it's in the latest `1.0-SNAPSHOT`. Url for the snapshot repo is available in the Book of Geb.

topr:
@erdi 
I'm trying to use the snapshot and Grails is struggling with transitive dependency:
> Error Resolve error obtaining dependencies: Could not find artifact org.grails.plugins:geb:zip:1.0-SNAPSHOT

I'm wondering, should I just use a released version of the Geb plugin, just try to explicit override its dependencies (the geb-core, etc...) or is there any other maven repo I should add to the build config?

At the moment I have these two added apart from a few other which has been in the project before:
 - https://oss.sonatype.org/content/repositories/releases/
 - https://repo.grails.org/grails/snapshots/

I'm able to see the relevant version of the Geb plugin over here:
https://repo.grails.org/grails/snapshots/org/grails/plugins/geb/1.0-SNAPSHOT/
even though it's not being fetched (there is no zip?)?

By the way, why there are two Geb snapshots: 1.0-SNAPSHOT and 1.0.0-SNAPSHOT. Which one is the right one? Latter is more recent but doesn't seem to contain the change from this PR.

erdi:
I assume you're using Grails 2.x, right? We don't release snapshot versions of Grails plugin so the best thing to do is to use the latest released version of the plugin, exclude all Geb dependencies it pulls in and then declare dependency on `org.gebish:geb-spock:1.0-SNAPSHOT` (assuming you're using Geb with spock).

If you are using Geb with Grails 3.x then the plugin is provided by the Grails team and has a Geb version hardcoded to `0.12.2` of the `1.0.1` version of the plugin. You will need to do the same dance with excluding and including or use a [version replacement](https://docs.gradle.org/current/userguide/dependency_management.html#sec:blacklisting_version):

    configurations.all {
        resolutionStrategy.eachDependency { details ->
            if (details.requested.group == "org.gebish") {
                details.useVersion "1.0-SNAPSHOT"
            }
        }
    }

As I've mentioned before, Geb snapshots are available from a [repo which is given in the Book of Geb](http://www.gebish.org/manual/current/#snapshot-repository).

topr:
Thanks, this was indeed for Grails 2.x.
Leaving the released version of Grails plugin while using geb-spock in the SNAPSHOT version worked well. I just didn't want to mess up dependencies too much if it was only repos misconfiguration.

Do you have any idea what's up with these SNAPSHOT versions of Geb, why there is 1.0-SNAPSHOT (April) and 1.0.0-SNAPSHOT (more recent). It seems that there have been no changes since April anyway. Where does the newer SNAPSHOT comes from and which one is the right one to use (I mean which one is going to become released version)?

erdi:
1.0.0-SNAPSHOT is an old version, which I don't use anymore. I decided to release minor versions without patch version path so 1.0-SNAPSHOT is the latest one. 

I did not push any changes since late April hence the latest build puished this pom: https://oss.sonatype.org/content/repositories/snapshots/org/gebish/geb-core/1.0-SNAPSHOT/geb-core-1.0-20160421.093528-29.pom.sha1.

PR #123, opened 2016-10-18
Move report label creation into its own method so subclasses can generate report labels using the correct test and report numbers

craigatk:
I'm working on a subclass of `GebReportingSpec` to be able to use multiple browsers in a single Geb test for verifying things like cross-browser WebSocket communication. (https://github.com/craigatk/ratpack-standup/blob/master/src/test/groovy/standup/geb/multi/MultiBrowserGebSpec.groovy)

In my `MultiBrowserGebSpec` sublcass I have multiple `Browser`instances that I want to write reports for, but I can't get access to the `gebReportingSpecTestCounter` and `gebReportingPerTestCounter` fields to create the report label because those fields are private. (https://github.com/craigatk/ratpack-standup/blob/master/src/test/groovy/standup/geb/multi/MultiBrowserGebSpec.groovy#L80)

I suggest moving the report label into its own method so subclasses can generate report labels using the correct test and report numbers. If there is a better way to do this, just let me know. Thanks!

### DIFF
--- a/module/geb-spock/src/test/groovy/geb/spock/ReportNameGebSpec.groovy
+++ b/module/geb-spock/src/test/groovy/geb/spock/ReportNameGebSpec.groovy
@@ -1,0 +1,26 @@
+/* Copyright 2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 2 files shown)


erdi:
LGTM.

I just have one request - can we please add a test exercising this method directly to make it explicitly part of the public API and guard it from being inadvertently removed during a refactoring?

craigatk:
Good idea, I'll add a test. Thanks Marcin!

module/geb-spock/src/test/groovy/geb/spock/ReportNameGebSpec.groovy
```diff
@@ -0,0 +1,37 @@
...
+class ReportNameGebSpec extends GebReportingSpec {
+    @Shared
+    TestHttpServer server = new CallbackHttpServer()
+
+    def setupSpec() {
+        server.start()
+    }
+
+    def setup() {
+        baseUrl = server.baseUrl
```
erdi: None of the above setup will be necessary if you set the `reportOnTestFailureOnly` config option to true.

craigatk: Test updated.

erdi:
Thanks for the contribution, Craig.

PR #124, opened 2016-10-29
Let gebish.org/manual/ redirect to current manual

JacobAae:
A possible fix for https://github.com/geb/issues/issues/415
Instead of a 404, we could let it redirect to the current manual, which seems to be a fair default

### DIFF
--- a/doc/site/src/ratpack/Ratpack.groovy
+++ b/doc/site/src/ratpack/Ratpack.groovy
@@ -33,6 +33,9 @@ ratpack {
         add Date, new Date()
     }
     handlers {
+        get('manual') {
+            redirect("/manual/current")
+        }
         files {
             dir("public")
             indexFiles("index.html")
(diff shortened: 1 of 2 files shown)


erdi:
Sorry, but I still believe that it should be a 404 and not a redirect. Apart from that, your solution does not solve the problem in full because http://www.gebish.org/foo/ also serves an unstyled page instead of a 404. 

We would need a generic solution for this and a test as well, please. :)

JacobAae:
The trailing slashforward causes the static assets to be served from the wrong uri. This solution removes any trailing slashforward, and redirects to the page without it. This makes gebish.org/manual/ and other pages display the "Not Found" page. It does not actually serve a 404 status, but this is a general issue.

JacobAae:
Added 404 status if Not Found page is the one displayed

erdi:
Thanks yet again, @JacobAae!

PR #125, opened 2016-10-29
Added high z-index value on menu dropdown to display properly

JacobAae:
Should make dropdown menus show correctly, and take care of: https://github.com/geb/issues/issues/414


erdi:
Thanks Jacob.

PR #126, opened 2016-11-12
Docs: Clarify auto clearing cookies is only for current domain

pradtke:


erdi:
Thank you very much for the contribution, @pradtke. Documentation improvements are always highly appreciated.

PR #127, opened 2016-11-21
Fix for #450 unexpected pages are now checked after the atCheck fails

leonard84:


module/geb-core/src/main/groovy/geb/AtVerificationResult.groovy
```diff
@@ -34,6 +34,14 @@ class AtVerificationResult {
         errorThrown == null && atResult
     }
 
+    AssertionError getErrorThrown() {
+        errorThrown
+    }
+
+    boolean getAtResult() {
```
erdi: This method is unused and thus unnecessary.

module/geb-core/src/main/groovy/geb/Page.groovy
```diff
@@ -178,8 +178,14 @@ class Page implements Navigable, PageContentContainer, Initializable, WaitingSup
      * @throws UnexpectedPageException when at an unexpected page
      */
     boolean verifyAt() {
-        getInitializedBrowser().checkIfAtAnUnexpectedPage(getClass())
-        verifyThisPageAtOnly(true)
+        def verificationResult = getAtVerificationResult(true)
+        if (!verificationResult) {
+            getInitializedBrowser().checkIfAtAnUnexpectedPage(getClass())
+            if (verificationResult.errorThrown) {
```
erdi: Instead of exposing `errorThrown` property please add `AtVerificationResult.rethrowAnyErrors()` which will possibly throw the error if one has occurred and use it here.

module/geb-core/src/test/groovy/geb/UnexpectedPagesSpec.groovy
```diff
@@ -94,6 +94,30 @@ class UnexpectedPagesSpec extends GebSpecWithCallbackServer {
...
+        defineUnexpectedPages(ConflictingUnexpectedPage)
+
+        when:
+        via ExpectedPage
+
+        then:
+        at ExpectedPage
+    }
+
+    void 'unexpected pages are checked only if the atCheck for the expected page fails'() {
```
erdi: I think that this test is redundant because it tests the exact same path as the one which follows, i.e. `an exception is thrown when we end up at an unexpected page`. Or am I'm missing something?

erdi:
I applied the changes previously requested myself. Thanks for the contribution @leonard84!

PR #128, opened 2017-03-29
Patch for https://github.com/geb/issues/issues/477

mahileeb:
If select is set to null, deselect all if a multi select or throw an exception if not multiple


antony:
Hi @mahileeb 

Unfortunately your commit fails CodeNarc validation - see the failed build link above for more details on this.

Also - and especially in the case of a testing framework, pull requests should include accompanying tests - ones which would fail prior to the code change, to both demonstrate the purpose of the fix, and to ensure that the fix actually works.

mahileeb:
Sorry, fired this off quickly - will look at the failures and resubmit with a test

mahileeb:
Hi @antony, I've fixed up code narc failures and added test cases for single and multi select cases

erdi:
Thanks for the contribution, @mahileeb. FWIW I reworked it a bit: https://github.com/geb/geb/commit/72b1d3bc910c612a6611ae242cf98be20db8ceed.

mahileeb:
Thanks @erdi!

PR #130, opened 2017-05-05
Navigator.filter is made case insensitive

rickirunge:
Solves problem described by issues:
https://github.com/geb/issues/issues/482


erdi:
Thanks for the contribution @rickirunge!

PR #136, opened 2018-12-07
Add missing text-like input types

mkutz:
See geb/issues#548

### DIFF
--- a/module/geb-core/src/main/groovy/geb/module/NumberInput.groovy
+++ b/module/geb-core/src/main/groovy/geb/module/NumberInput.groovy
@@ -15,8 +15,6 @@
  */
 package geb.module
 
-import javax.annotation.Nonnull
-
 class NumberInput extends AbstractInput {
 
     final String inputType = "number"
@@ -31,30 +29,28 @@ class NumberInput extends AbstractInput {
     }
 
     Number getMin() {
-        String minString = attr("min")
-        minString ? parseNumber(minString) : null
+        parseNumber(attr("min"))
     }
 
     Number getMax() {
-        String maxString = attr("max")
-        maxString ? parseNumber(maxString) : null
+        parseNumber(attr("max"))
     }
 
     Number getStep() {
-        String stepString = attr("step")
-        stepString ? parseNumber(stepString) : 1
+        parseNumber(attr("step"))
     }
 
-    private static Number parseNumber(@Nonnull String string) {
-        if (!string?.number) {
-            throw new NumberFormatException("Value \"${string}\" of ${this} is not a number.")
+    private Number parseNumber(String string) {
+        if (null == string) {
+            return null
         }
-        if (string?.bigDecimal) {
+        if (string.bigDecimal) {
             return string.toBigDecimal()
         }
-        if (string?.bigInteger) {
-            return string.toBigInteger()
+        if (string.integer) {
+            return string.toInteger()
         }
+        throw new NumberFormatException("Value \"${string}\" of ${this} is not a number.")
     }
 
 }
--- a/module/geb-core/src/main/groovy/geb/module/NumberInput.groovy
+++ b/module/geb-core/src/main/groovy/geb/module/NumberInput.groovy
@@ -41,7 +41,7 @@ class NumberInput extends AbstractInput {
     }
 
     private Number parseNumber(String string) {
-        if (null == string) {
+        if (string == null) {
             return null
         }
         if (string.bigDecimal) {
--- a/module/geb-core/src/main/groovy/geb/module/NumberInput.groovy
+++ b/module/geb-core/src/main/groovy/geb/module/NumberInput.groovy
@@ -15,27 +15,46 @@
  */
 package geb.module
 
-class NumberInput extends TextLikeInput {
+import javax.annotation.Nonnull
+
+class NumberInput extends AbstractInput {
 
     final String inputType = "number"
 
-    void setNumber(int number) {
+    void setNumber(Number number) {
         value(number)
     }
 
-    Integer getNumber() {
+    Number getNumber() {
         String value = value()
-        value ? Integer.valueOf(value) : null
+        value ? parseNumber(value) : null
     }
 
-    Integer getMin() {
+    Number getMin() {
         String minString = attr("min")
-        minString ? Integer.valueOf(minString) : null
+        minString ? parseNumber(minString) : null
     }
 
-    Integer getMax() {
+    Number getMax() {
         String maxString = attr("max")
-        maxString ? Integer.valueOf(maxString) : null
+        maxString ? parseNumber(maxString) : null
+    }
+
+    Number getStep() {
+        String stepString = attr("step")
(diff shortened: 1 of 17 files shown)


erdi:
Looks like [`:doc:manual-snippets:codenarcTest` is failing](https://circleci.com/gh/geb/geb/799?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) for this PR. Can you please fix?

mkutz:
WaitingSupportSpec > larger interval than timeout throwing exception fails, but I cannot see any connection to my changes.

erdi:
Yeah, probably unrelated. I'll see if I can make it less flakey.

mkutz:
Sorry, that last commit wasn't meant to go into this PR.

erdi:
That’s fine. You can always reset the branch to get rid of the last two commits from the history completely and then force push.

mkutz:
Reading the specs of W3C, I just realized that `NumberInput` is supposed to allow float values, only the default has a `step=1`. I'll fix that on Monday.

erdi:
I had a look at this and it support for `url` and `password` input types are spot on. Thanks for also including changes to the docs.

With regards to support for the `number` input type I'd argue that it should not extend `TextLikeInput` but extend `AbstractInput` like `DateInput` for example does. The value of that input type is typed (it's a number) so having a `text` property makes little sense for that type. I'd like to keep things consistent with other types that have typed values, e.g. aforementioned `DateInput`.

You've noted that this particular input type deals with both integer and floating point numbers. Let's therefore change the type of `number` property on `NumberInput` to `java.lang.Number` and return an `Integer` if there are no decimal places in the value or `BigDecimal` if there are.

Finally, please add yourself to the [contributor list in the manual](https://github.com/geb/geb/blob/master/doc/manual/src/docs/asciidoc/140-project.adoc#contributors) as part of this PR.

mkutz:
> I had a look at this and it support for `url` and `password` input types are spot on. Thanks for also including changes to the docs.
> 
> With regards to support for the `number` input type I'd argue that it should not extend `TextLikeInput` but extend `AbstractInput` like `DateInput` for example does. The value of that input type is typed (it's a number) so having a `text` property makes little sense for that type. I'd like to keep things consistent with other types that have typed values, e.g. aforementioned `DateInput`.

OK, makes sense.

> You've noted that this particular input type deals with both integer and floating point numbers. Let's therefore change the type of `number` property on `NumberInput` to `java.lang.Number` and return an `Integer` if there are no decimal places in the value or `BigDecimal` if there are.

I thought about `Number` as well, but rather `BigInteger` and `BigDecimal` to be more tolerant. What do you think?

How do you feel about the `getMin` and `getMax` methods I mixed in? I'm thinking about either kick them out of put in `getStep`.

> Finally, please add yourself to the [contributor list in the manual](https://github.com/geb/geb/blob/master/doc/manual/src/docs/asciidoc/140-project.adoc#contributors) as part of this PR.

Will do :)

erdi:
> I thought about Number as well, but rather BigInteger and BigDecimal to be more tolerant. What do you think?

Can you please elaborate on why you would like to use `BigInteger`? I would find it surprising as a user to get a `BigInteger` and not `Integer`. According to [Groovy documentation](http://groovy-lang.org/differences.html#_conversions) it should be cast to `Integer` on the fly, but still...

> How do you feel about the getMin and getMax methods I mixed in? I'm thinking about either kick them out of put in getStep.

I think some people might find it useful. Especially if they will deal with `Number`s. I agree with either adding `getStep()` or removing them completely. It's up to you.

module/geb-core/src/main/groovy/geb/module/NumberInput.groovy
```diff
@@ -0,0 +1,60 @@
...
+        String maxString = attr("max")
+        maxString ? parseNumber(maxString) : null
+    }
+
+    Number getStep() {
+        String stepString = attr("step")
+        stepString ? parseNumber(stepString) : 1
+    }
+
+    private static Number parseNumber(@Nonnull String string) {
```
mkutz: @erdi, is there a place where to put this method? Do you know a better implementation?

module/geb-core/src/main/groovy/geb/module/NumberInput.groovy
```diff
@@ -0,0 +1,60 @@
...
+        if (!string?.number) {
+            throw new NumberFormatException("Value \"${string}\" of ${this} is not a number.")
+        }
+        if (string?.bigDecimal) {
+            return string.toBigDecimal()
+        }
+        if (string?.bigInteger) {
```
mkutz: I think `BigInteger` is safer then `Integer`, but it's your decision, @erdi.

erdi: Let's use `Integer` here, please.

module/geb-core/src/main/groovy/geb/module/NumberInput.groovy
erdi: Nope, I'd keep the method where it is. Just please don't use `@Nonnull`, it's not something that is used in this particular codebase.

module/geb-core/src/main/groovy/geb/module/NumberInput.groovy
```diff
@@ -0,0 +1,60 @@
...
+    void setNumber(Number number) {
+        value(number)
+    }
+
+    Number getNumber() {
+        String value = value()
+        value ? parseNumber(value) : null
+    }
+
+    Number getMin() {
```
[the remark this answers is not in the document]
mkutz: Alright. I only returned 1 for step since in the [W3C spec](https://www.w3.org/TR/2017/REC-html52-20171214/sec-forms.html#number-state-typenumber) it is documented as the default step. So from a less technical point of view returning 1 is correct if its not set. But I also think that we should stick to the technical interface-to-HTML point of view.

module/geb-core/src/main/groovy/geb/module/NumberInput.groovy
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    Number getStep() {
+        String stepString = attr("step")
+        stepString ? parseNumber(stepString) : 1
+    }
+
+    private static Number parseNumber(@Nonnull String string) {
+        if (!string?.number) {
+            throw new NumberFormatException("Value \"${string}\" of ${this} is not a number.")
```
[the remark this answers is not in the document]
mkutz: Oh, my bad. Thanks.

erdi:
Awesome, high quality contribution, @mkutz 👍. Hopefully not the last one 😃! Thanks!

mkutz:
You're welcome 😊, thanks. Took more time than I expected but was also kind of fun. I'll definitely will take care of the other input types in the next couple of weeks.

erdi:
It always takes more time than we expect, it’s software development in the
end ;).

BTW, I’m planning to do another relese before the end of the year so any
PRs made before then will be part of it.

PR #138, opened 2018-12-13
Add color input module

mkutz:
See geb/issues#549


mkutz:
Just ran that test 200 times with no failure.

erdi:
As long as the build passes on CircleCI I'm happy. The Windows build is there only because we had tests that were not cross platform in the past and would fail on that OS and the failure is definitely not Windows specific and has nothing to do with your changes.

PR #141, opened 2018-12-17
Add time input module

mkutz:
See geb/issues#554


mkutz:
I screwed up again. Trying to fix. The PR should only have one single commit and no conflicts at all, just picked the wrong branch.

PR #148, opened 2019-03-22
Use Number instead of Double in Wait

mkutz:
There is no reason to restrict timeout and interval to Double values.
In the documentation there never was any example using Double, but
Integer (e.g. 5) or Float (e.g. 0.1). While the calls actually worked
due to Groovy's dynamic typing, they cause errors when using strict
typing and cause irritations in Idea, which cannot find the right
method and assumes that waitFor will return Object instead of the return
type of the given Closure.

Switching to Number allows to use any Number type and will solve these
issues.

Contributes to geb/issues#570

### DIFF
--- a/doc/manual/src/docs/asciidoc/140-project.adoc
+++ b/doc/manual/src/docs/asciidoc/140-project.adoc
@@ -86,7 +86,7 @@ This page lists the high level changes between versions of Geb.
 
 === {geb-version}
 
-No changes have been made, yet.
+* Use `Number` instead of `Double` for waitFor calls [issue:570[]]
 
 === 2.3.1
 
--- a/module/geb-waiting/src/main/groovy/geb/waiting/Wait.groovy
+++ b/module/geb-waiting/src/main/groovy/geb/waiting/Wait.groovy
@@ -26,24 +26,24 @@ class Wait {
     /**
      * 5 seconds
      */
-    static public final Double DEFAULT_TIMEOUT = 5
+    static public final Number DEFAULT_TIMEOUT = 5
 
     /**
      * 100 milliseconds
      */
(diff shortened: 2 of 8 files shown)


doc/manual/src/docs/asciidoc/140-project.adoc
```diff
@@ -86,7 +86,7 @@ This page lists the high level changes between versions of Geb.
 
 === {geb-version}
 
-No changes have been made, yet.
+* Use `Number` instead of `Double` for waitFor calls [issue:570[]]
```
erdi: 👍

module/geb-waiting/src/main/groovy/geb/waiting/Wait.groovy
```diff
@@ -141,10 +141,10 @@ class Wait {
      * Blocks the caller for the retryInterval
      */
     void sleepForRetryInterval() {
-        Thread.sleep(toMiliseconds(retryInterval) as long)
+        Thread.sleep(toMilliseconds(retryInterval) as long)
     }
 
-    private double toMiliseconds(Double seconds) {
+    private static double toMilliseconds(Number seconds) {
```
erdi: I think that if we change the return type of this method to `int` then we won't need the `Math.ceil` call and a cast where this method is used.

mkutz: If I understand your suggestion correctly, it's not exactly the same sind we'd change ceil to floor. But I guess nobody will care about the milli fractions.

Returning int is definitely a good idea. I'll take care of it tomorrow.

erdi:
Thank you!

PR #149, opened 2019-04-12
Remove @RequiresRealBrowser

mkutz:
Since https://sourceforge.net/p/htmlunit/bugs/1923/ is now closed, this
test now works with HtmlUnit.


erdi:
Thanks.

What about other tests that you added and which are annotated with:
```
@RequiresRealBrowser // maybe due to https://sourceforge.net/p/htmlunit/bugs/1923/
```
?

mkutz:
I tried only one till now and it was still failing. I will check next week but I guess they are all HtmlUnit issues in their own right.

erdi:
Understood.

PR #153, opened 2019-08-28
events tracking

darionct:
Please review the pull request and add comments and suggestions.
Thanks!


erdi:
Thanks for the PR, @darionct. Unfortunately, after looking deeper into this, I came to a realisation that there is certain complexity involved around passing the event source to the listener. So, if a module instance is clicked we want to pass that module instance as the source and not the underpinning navigator to which the module instance delegates as is the case in your implementation. Same goes for clicking on page content. 

I have an idea how to implement it but it's quite complex and hard to explain. I will prioritise working on this and cut a release as soon as I'm done. Hopefully you're ok with it.

PR #154, opened 2019-11-12
Added pom type for groovy-all dependency in manual

tombrown1002:
No longer downloads as a jar, so addition of pom type necessary.


erdi:
Thanks for the contribution, @tombrown1002 but the explicit dependency on groovy is no longer necessary because [it's already included in Geb's pom](https://search.maven.org/remotecontent?filepath=org/gebish/geb-core/3.2/geb-core-3.2.pom). Would you mind removing it instead of modifying it? Also, can you please [add yourself as a contributor](https://github.com/geb/geb/blob/master/doc/manual/src/docs/asciidoc/140-project.adoc#contributors) while you're at it?

erdi:
I've applied 86f28b5d instead.

PR #151, opened 2019-04-30
Replace link to old selenium wiki in 010-intro.adoc

Topperfalkon:
Selenium wiki was migrated from google code to github, so the link redirects, but not to the correct page. I've replaced the link with a direct link to the corresponding page on the github wiki

PR #157, opened 2020-02-20
Update `Page Objects` url to its new location

jrodalo:
This was fixed in the docs (#151) but the main site (http://www.gebish.org) is still pointing to "code.google.com" which then redirects to an archived repo: https://github.com/seleniumhq/selenium-google-code-issue-archive


erdi:
Thanks for the contribution, @jrodalo.

PR #164, opened 2020-07-02
Use specificationContext to name reports

mkutz:
As Spock 2.0 is based on JUnit 5, `@Rule`s won't work anymore.

Using Spock's ISpecificationContext works on both current versions of
Spock, 1.3 and 2.0-M1.

### DIFF
--- a/module/geb-spock/src/main/groovy/geb/spock/GebReportingSpec.groovy
+++ b/module/geb-spock/src/main/groovy/geb/spock/GebReportingSpec.groovy
@@ -21,9 +21,6 @@ import spock.lang.Shared
 
 class GebReportingSpec extends GebSpec {
 
-    // Ridiculous name to avoid name clashes
-    @Rule
-    TestName gebReportingSpecTestName
     private int gebReportingPerTestCounter = 1
     @Shared
     private int gebReportingSpecTestCounter = 1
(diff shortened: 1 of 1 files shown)


erdi:
Thanks for this, @mkutz, this is a very good find! I don't plan doing any more 3.x releases and started the [v4.0 branch](https://github.com/geb/geb/tree/v4.0). The testing integration has been significantly reworked on it with the aim to deduplicate it and make it reusable (which was long overdue) so that it's easier to add support for additional testing frameworks (not only to the core but also by users) in preparation for adding support for JUnit 5. Would you mind repointing this PR to that branch? I can also just apply the change to that branch manually, just please let me know.

mkutz:
I started to do it, but as far as I can see, you are still relying on `org.junit.rules.TestRule` and –if I'm not mistaken– Spock 2.0 will remove the JUnit 4 dependency. We could replace that with an JUnit 5 extension, but that would be a breaking change for people sticking to JUnit 4 and/or Spock 1.x.

I'll try to get rid of all the JUnit 4 stuff in the PR, if you're OK with that.

mkutz:
Oh, never mind my last comment. Found the line I was looking for.

mkutz:
Closing this PR, will open another one for v4.0 branch.

erdi:
Thanks!

PR #165, opened 2020-07-03
Use specificationContext to name reports

mkutz:
As Spock 2.0 is based on JUnit 5, `@Rule`s won't work anymore.

Using Spock's ISpecificationContext works on both current versions of
Spock, 1.3 and 2.0-M1.

### DIFF
--- a/module/geb-spock/src/main/groovy/geb/spock/GebSpec.groovy
+++ b/module/geb-spock/src/main/groovy/geb/spock/GebSpec.groovy
@@ -25,9 +25,6 @@ class GebSpec extends Specification {
 
     private final static GebTestManager TEST_MANAGER = new SpockGebTestManagerBuilder().build()
 
-    @Rule
-    TestName gebSpecTestName
-
     @Delegate(includes = ["getBrowser"])
     GebTestManager getTestManager() {
         TEST_MANAGER
@@ -38,7 +35,7 @@ class GebSpec extends Specification {
     }
 
     def setup() {
-        testManager.beforeTest(gebSpecTestName.methodName)
+        testManager.beforeTest(specificationContext?.currentIteration?.name)
     }
 
     def cleanup() {


module/geb-spock/src/main/groovy/geb/spock/GebSpec.groovy
```diff
@@ -38,7 +35,7 @@ class GebSpec extends Specification {
     }
 
     def setup() {
-        testManager.beforeTest(gebSpecTestName.methodName)
+        testManager.beforeTest(specificationContext?.currentIteration?.name)
```
erdi, line 38: What's the situation in which `specificationContext` would be `null` or `currentIteration` would be `null` in this line? I would like us not to use the null-safe operator unless there is a reason for that. Feels too me like overly defensive programming, to be honest.

module/geb-spock/src/main/groovy/geb/spock/GebSpec.groovy
```diff
@@ -25,9 +25,6 @@ class GebSpec extends Specification {
 
     private final static GebTestManager TEST_MANAGER = new SpockGebTestManagerBuilder().build()
 
-    @Rule
-    TestName gebSpecTestName
```
erdi, line 29: The removal of this line without removing the import causes a codenarc violation and the build is failing. Please fix.

module/geb-spock/src/main/groovy/geb/spock/GebSpec.groovy
mkutz: You're right. I was inspired by the original line, but this indeed does not require being that defensive.

erdi:
I decided to pull and fix the issues myself, hopefully that's ok with you, @mkutz. :)

mkutz:
Sorry, didn't pay too much attention to this lately. Go ahead, @erdi

PR #167, opened 2020-12-11
Add pronunciation hint + original meaning of "Geb"

kriegaex:
The introduction in the Boof of Geb now reads:

> Geb (pronounced "jeb" and originally a contraction of "Groovy web") is a
> developer focused tool (...)


erdi:
Thanks for the contribution and apologies for the delay in merging - I was getting 4.0 out of the door.

PR #169, opened 2021-03-11
Fix Gradle deprecation warnings

pbielicki:
Before: https://scans.gradle.com/s/ijfek46ea7ecm/deprecations
After: https://scans.gradle.com/s/wq5mwytaylkcs/deprecations

### DIFF
--- a/internal/test-support-spock/test-support-spock.gradle
+++ b/internal/test-support-spock/test-support-spock.gradle
@@ -1,4 +1,9 @@
+plugins {
+	id 'java-library'
+}
+
 dependencies {
-	compile project(':internal:test-support'),
-		project(':module:geb-spock')
+	api project(':internal:test-support')
+	implementation project(':module:geb-ast')
+	implementation project(':module:geb-spock')
 }
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -22,10 +22,10 @@ repositories {
 }
 
 dependencies {
-	compile localGroovy()
-	compile gradleApi()
+	implementation localGroovy()
+	implementation gradleApi()
 
-	compile "com.google.guava:guava:23.0"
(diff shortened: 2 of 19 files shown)


erdi:
Would you mind rebasing this onto latest `master`, please, @pbielicki? I inadvertently [broke PR builds on CI when I enabled build cache](https://app.circleci.com/pipelines/github/geb/geb/105/workflows/339154fd-611a-4a13-bcae-a6ea0e33afc7/jobs/3653) and I've just got it fixed now.

internal/test-support-spock/test-support-spock.gradle
```diff
@@ -1,4 +1,9 @@
+plugins {
```
erdi, line 1: I hope it's ok with you but I think I will remove all applications of `java-library` plugin and just stick to `implementation` configuration for now. I think more thought needs to be put into what should go into `api` configurations and from what I've seen the `java-library` plugin has not been applied by you to all the places where it potentially could be. I've seen the Groovy compiler loading classes during compilation which seemingly should not have been loaded so for now, until I have the time to properly look into this, I'd rather we kept adding too many dependencies to consumer's compilation classpath than them getting surprising class loading exceptions during compilation if we get it wrong. 

I have created an issue to track that for the future: https://github.com/geb/issues/issues/636

erdi: Hmmm.... It looks like I got it wrong and it's the other way around - I need the `java-library` plugin applied and all of the dependencies need to go into `api` if we want to play it safe for now.

erdi:
Thanks for the contribution. I tweaked it a bit before merging in but it was still very helpful and saved me time.

pbielicki:
@erdi thanks for adapting and merging - I was sick on Friday and wasn't able to take a look

🎉

PR #173, opened 2021-11-10
Bump jodd core and lagarto versions

arixmkii:
Bumping `lagarto` because older versions depends on `jodd-log`, which depends on really narrow range of `log4j [2.8, 2.9)` and breaks some builds, where `log4j2` libraries are used.

As `lagarto` no longer depends on core, I had to add `jodd-core` dependency separately.

`jodd-core` is needed for
```groovy
import jodd.util.collection.CompositeIterator
```


erdi:
Thanks for the contribution and clear explanation for submitting. I will have time to look at this on Friday. In the meantime, can you please add yourself to [the list of contributors](https://github.com/geb/geb/blob/master/doc/manual/src/docs/asciidoc/140-project.adoc#contributors) as part of this PR?

arixmkii:
Updated the list.

erdi:
Thanks again for the contribution!

PR #176, opened 2022-03-25
Use proper Spock extension points

Vampire:
This change improves the used Spock extension points to use more appropriate ones.

### DIFF
--- a/module/geb-core/src/test/groovy/geb/ContentWaitConditionSpec.groovy
+++ b/module/geb-core/src/test/groovy/geb/ContentWaitConditionSpec.groovy
@@ -56,7 +56,7 @@ class ContentWaitConditionSpec extends GebSpecWithCallbackServer {
 
         then:
         WaitTimeoutException exception = thrown()
-        exception.cause.message.contains("it.displayed")
+        exception.cause.message.contains("it.focused")
     }
 
 }
@@ -64,6 +64,6 @@ class ContentWaitConditionSpec extends GebSpecWithCallbackServer {
 class ContentWaitConditionSpecPage extends Page {
     static content = {
         div(waitCondition: { it.displayed }) { $('div') }
-        failingDiv(wait: 0.05, waitCondition: { it.displayed }) { $('div') }
+        failingDiv(wait: 0.05, waitCondition: { it.focused }) { $('div') }
     }
 }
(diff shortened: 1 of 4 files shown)


module/geb-spock/src/main/groovy/geb/spock/GebExtension.groovy
```diff
@@ -38,42 +38,34 @@ class GebExtension implements IGlobalExtension {
         }
     }
 
-    protected GebTestManager getManager(IMethodInvocation invocation) {
+    protected static GebTestManager getManager(IMethodInvocation invocation) {
```
erdi, line 41: Looks good, thanks. 

One thing, though - please do not make these methods static. I know that IntelliJ says that they can be static but the convention on the project is that methods are only static if they have to, i.e. they are called from a static context and not if they can be static because they don't use any instance level fields or methods.

module/geb-core/src/test/groovy/geb/waiting/WaitingContentSpec.groovy
```diff
@@ -43,6 +43,9 @@ class WaitingContentSpec extends WaitingSpec {
     }
 
     def "wait timeout"() {
+        given:
+        showDelay = 10
```
erdi, line 47: This is good, thanks! It's not always obvious how to make tests for `Wait` fast, i.e. waiting as little as possible while still keeping them reliable.

module/geb-core/src/test/groovy/geb/ContentWaitConditionSpec.groovy
```diff
@@ -56,14 +56,14 @@ class ContentWaitConditionSpec extends GebSpecWithCallbackServer {
...
+        exception.cause.message.contains("it.focused")
     }
 
 }
 
 class ContentWaitConditionSpecPage extends Page {
     static content = {
         div(waitCondition: { it.displayed }) { $('div') }
-        failingDiv(wait: 0.05, waitCondition: { it.displayed }) { $('div') }
+        failingDiv(wait: 0.05, waitCondition: { it.focused }) { $('div') }
```
erdi, line 67: Just use a condition that will always fail irrespective of any timing conditions, nice. 👍

module/geb-spock/src/main/groovy/geb/spock/GebExtension.groovy
Vampire: Ah, sorry, didn't know that. Fix is pushed.

erdi: Cool. Will merge if all checks pass.

PR #178, opened 2022-03-27
Use System.nanoTime for duration measurement

Vampire:
For measuring durations within one JVM instance, the most appropriate method is `System.nanoTime()` as it is independent of any wall clocks, DST changes and so on.

### DIFF
--- a/module/geb-waiting/src/main/groovy/geb/waiting/Wait.groovy
+++ b/module/geb-waiting/src/main/groovy/geb/waiting/Wait.groovy
@@ -15,7 +15,7 @@
  */
 package geb.waiting
 
-import java.time.Instant
+import static java.util.concurrent.TimeUnit.MILLISECONDS
 
 /**
  * Represents a particular configuration of waiting, but does not encompass what is to be waited on.
@@ -107,7 +107,7 @@ class Wait {
             thrown = e
         }
 
-        def timedOut = Instant.now() > timeoutThreshold
+        def timedOut = timeoutThreshold - System.nanoTime() <= 0
         while (!pass && !timedOut) {
             sleepForRetryInterval()
             try {
@@ -117,7 +117,7 @@ class Wait {
                 pass = new UnknownWaitForEvaluationResult(e)
                 thrown = e
             } finally {
-                timedOut = Instant.now() > timeoutThreshold
+                timedOut = timeoutThreshold - System.nanoTime() <= 0
             }
         }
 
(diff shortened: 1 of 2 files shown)


erdi:
I'm a bit weary of this change. I don't mind changing the test although I don't think the code there is incorrect in any way, but I don't want to change prod code unless there is an actual problem that your change fixes. How is the code in either of the places dependent of wall clocks or DST changes if it's using `Instant.now()`? As far as I can tell neither `Instant` nor `Duration` does in any shape or form depend on time zones... Can you please give a concrete example of when the current code does not correctly calculate elapsed time and yours does?

Vampire:
I said "wall clocks, DST changes and so on".
You are right that Instant of course is totally independent of DST, but no of wall clock.

With this code:
```kts
val start = Instant.now()
Thread.sleep(10_000)
println(Duration.between(start, Instant.now()))
Thread.sleep(10_000)
println(Duration.between(start, Instant.now()))
```
you get roughly
```
PT10.0129939S
PT20.0234877S
```
Now with the same code, within the second 10 seconds after the first line was printed, change the system time one hour forward and you get roughly
```
PT10.3247512S
PT1H20.3364294S
```

`System.nanoTime()` is designed exactly for the use-case of reliably measuring a duration within one JVM without depending on any "natural" time.

erdi:
Right. So the potential problem this solves is if the system time is changed (and not because of DST or anything but only if somebody changes the offset from epoch) when the method executes. I find it to be quite an obscure edge case to be honest and the code change required makes it slightly harder to understand, at least for me, but given you have already contributed this let's run with it.

module/geb-waiting/src/main/groovy/geb/waiting/Wait.groovy
```diff
@@ -107,7 +107,7 @@ class Wait {
             thrown = e
         }
 
-        def timedOut = Instant.now() > timeoutThreshold
+        def timedOut = timeoutThreshold - System.nanoTime() <= 0
```
erdi, line 110: FYI, I will change this to `System.nanoTime() > timeoutThreshold` after merging, I find that more readable

Vampire: @erdi me too, but you should not change it. The JavaDoc of `nanoTime` explicitly says

> one should use t1 - t0 < 0, not t1 < t0, because of the possibility of numerical overflow.

The first value you get can for example be `Long.MAX_VALUE - 10` and 20 nanoseconds later you get `Long.MIN_VALUE + 10`. The difference against 0 comparison still works as designed, the simple comparison not.

erdi: Uh, oh, so there is a reason. Makes sense, thanks for explaining. Off to revert the change, then.

PR #179, opened 2022-03-27
Generate automatic failure report for cleanupSpec error in correct report group

Vampire:
Currently a browser is instantiated early in `beforeTestClass` if reporting is enabled just to configure the reporting.
But if you have a mix of test that use Geb and test that don't use it, this is wasted effort.
Especially if you do more when a browser is instaniated automatically like auto-login into the SUT.

With this change the browser creator is decorated to do the additional actions,
so the browser is only created when it is actually used in the test.

### DIFF
--- a/module/geb-core/src/main/groovy/geb/test/GebTestManager.groovy
+++ b/module/geb-core/src/main/groovy/geb/test/GebTestManager.groovy
@@ -19,6 +19,7 @@ import geb.Browser
 
 import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicInteger
+import java.util.function.Consumer
 import java.util.function.Predicate
 import java.util.function.Supplier
 
@@ -29,6 +30,7 @@ class GebTestManager {
     private final static Map<Class<?>, AtomicInteger> TEST_COUNTERS = new ConcurrentHashMap<>()
 
     private final Supplier<Browser> browserCreator
+    private Consumer<Browser> browserConfigurer = {}
     private final Predicate<Class<?>> resetBrowserAfterEachTestPredicate
     final boolean reportingEnabled
 
@@ -73,6 +75,9 @@ class GebTestManager {
         if (reportingEnabled) {
             getBrowser().reportGroup(testClass)
             getBrowser().cleanReportGroupDir()
+            browserConfigurer = { Browser browser ->
+                browser.reportGroup(testClass)
+            }
             testCounter = nextTestCounter(currentTestClass)
             perTestReportCounter = 1
         }
@@ -82,7 +87,6 @@ class GebTestManager {
         currentTestClassStack.push(testClass)
         currentTestName = testName
         if (reportingEnabled) {
-            getBrowser().reportGroup(testClass)
             testCounter = nextTestCounter(currentTestClass)
             perTestReportCounter = 1
         }
@@ -106,6 +110,7 @@ class GebTestManager {
         if (!resetBrowserAfterEachTest) {
             resetBrowser()
         }
+        browserConfigurer = {}
         currentTestClassStack.pop()
     }
 
@@ -134,7 +139,8 @@ class GebTestManager {
     }
 
     private Browser createBrowser() {
-        browserCreator ? browserCreator.get() : new Browser()
+        (browserCreator ? browserCreator.get() : new Browser())
+                .tap(browserConfigurer.&accept)
     }
 
     private boolean getResetBrowserAfterEachTest() {
--- a/module/geb-junit4/src/test/groovy/geb/junit4/GebReportingTestTest.groovy
+++ b/module/geb-junit4/src/test/groovy/geb/junit4/GebReportingTestTest.groovy
@@ -69,7 +69,7 @@ class GebReportingTestTest extends GebReportingTest {
     def doTestReport() {
         if (++counter > 1) {
             def report = reportGroupDir.listFiles().find { it.name.startsWith("00" + (counter - 1)) && endsWith('end.html') }
-            assert report.exists()
+            assert report?.exists()
             assert report.text.contains('<div class="d1" id="d1">')
         }
     }
--- a/module/geb-spock/src/test/groovy/geb/spock/GebReportingSpecSpec.groovy
+++ b/module/geb-spock/src/test/groovy/geb/spock/GebReportingSpecSpec.groovy
@@ -277,7 +277,7 @@ class GebReportingSpecSpec extends Specification {
         """
 
         then:
-        new File(reportDir, "001-001-fixture-failure.html").exists()
+        reportFile("001-001-fixture-failure.html").exists()
     }
 
     SummarizedEngineExecutionResults runReportingSpec(String additionalConfiguration = "", String body) {


erdi:
> But if you have a mix of test that use Geb and test that don't use it, this is wasted effort.

In a single spec extending `GebReportingSpec`? Why would you ever have a test in a class extending base with reporting support if it does not do anything in a browser? Sorry, I'm struggling to find justification for such a setup. Your change makes the code much more complex in this area and I'd rather not do that unless it supports a sensible use case.

Vampire:
Well, it is mainly a suggestion, if you don't like it, you don't need to accept it.

I actually hit that case when I had that auto-login we talked about in Slack and then had another test feature in the same test class which does not use a browser and it was still created, popped up, logged in an all that just because Geb wanted to just set the reporting properties.

As this was easy to fix and in _my_ opinion didn't make the code that much more complex, I thought I send a PR with the changes.

One more thing to note is, that I think my changes actually also fix a bug.
If you look at the change in `GebReportingSpecSpec#failures in cleanupSpec methods are reported on`, you see that I changed the expectation.
Before, the `cleanupSpec` failure report was saved outside the report group.
With my changes it is where it belongs in my opinion, within the report group for that spec.

erdi:
I honestly don't think that it's worth adding this complexity to the code (even if it's not that big) to support a use case which can be solved in a simple and better way - if your test does not need `Browser` and reporting management then don't extend a base test class supplied by Geb.

With regards to the bug that you fixed, we could indeed keep that. But please do not make `browserCreator` mutable and instead introduce a separate mutable field which will hold a creator that decorates the browser returned from `browserCreator` by setting the reporting group on it. In `getBrowser()` we could use that new creator and default it in `beforeTestClass()` to `browserCreator` when reporting is disabled. We would keep the eager configuration of reporting in `beforeTestClass()`, though.

Vampire:
Your wish is my command

erdi:
Thanks but as you are probably already aware, the build looks quite badly broken...

module/geb-core/src/main/groovy/geb/test/GebTestManager.groovy
```diff
@@ -73,6 +77,9 @@ class GebTestManager {
         if (reportingEnabled) {
             getBrowser().reportGroup(testClass)
             getBrowser().cleanReportGroupDir()
+            browserConfigurer = { Browser browser ->
```
erdi, line 80: I would expect `browserConfigurer` to be cleared in `afterTestClass()`

erdi: Also, this should be set at the top of this `if` block and then `getBrowser().reportGroup(testClass)` should be removed

module/geb-core/src/main/groovy/geb/test/GebTestManager.groovy
```diff
@@ -49,7 +51,9 @@ class GebTestManager {
 
     Browser getBrowser() {
         if (browser == null) {
-            browser = createBrowser()
+            browser = createBrowser().tap {
```
erdi, line 54: I think it would be better for this change to go into `createBrowser()` and be simplified to `.tap(browserConfigurer.&accept)`

Vampire: That's where I would have put it too, but you said you want it in `getBrowser` :-D
I moved it now

module/geb-core/src/main/groovy/geb/test/GebTestManager.groovy
Vampire: > I would expect `browserConfigurer` to be cleared in `afterTestClass()`

I thought that's unnecessary as anyway each test class has its own test manager, but I added it now.

> Also, this should be set at the top of this `if` block and then `getBrowser().reportGroup(testClass)` should be removed

I thought about that, but it would change semantics.
The current code would also configure an already existing browser.
Like it is now, an existing and all future browsers are configured.

Vampire:
> Thanks but as you are probably already aware, the build looks quite badly broken...

Yes, I was aware of that, just didn't know why the hell it is happening, and too sleepy to investigate further.
The problem was the `createBrowser().tap { browserConfigurer.accept(it) }` where in the `tap` the `Browser` is the delegate and the resolve strategy is `DELEGATE_FIRST`. So `browserConfigurer` is frist searched in the browser, gives some exception that then somehow is just swallowed or whatever, causing that strange behavior.

Using a method reference or qualifying with `owner.browserConfigurer` or adding a local variable outside the closure that is then used or changing the resolution strategy to OWNER_FIRST or similar approaches to prevent the delegate being queried fixed the failing tests.

Vampire:
Btw. there is still a flak test in the AppVeyor build.
The build failed and I just retriggered it to become green.

Vampire:
https://ci.appveyor.com/project/erdi/geb/builds/43237499

I actually wonder how the content cannot be found after the at-check was successfully finding the content.

module/geb-core/src/main/groovy/geb/test/GebTestManager.groovy
erdi: Sorry, I wasn't aware that there is also `createBrowser()` when making that suggestion. Thanks for addressing it.

module/geb-core/src/main/groovy/geb/test/GebTestManager.groovy
erdi: > I thought that's unnecessary

It's indeed not strictly necessary but it feels cleaner for that field to not hold a reference to the class after `afterTestClass()` has been called.

> as anyway each test class has its own test manager

That's unfortunately not the case as instances of `GebTestManager` are referenced by static fields.

> I added it now.

Thanks.

> The current code would also configure an already existing browser.
> Like it is now, an existing and all future browsers are configured.

Hmm... That's a good catch. Ideally there should never be an existing browser. Any previous browser should be cleared by `afterTest()` or `afterTestClass()` but as things stand if you access `browser` property from `cleanupSpec()` in a non-stepwise spec then there will be one next time `beforeTestClass()` is called. Ok, please leave it as is to not inflate scope and I will tweak it later.

module/geb-core/src/main/groovy/geb/test/GebTestManager.groovy
```diff
@@ -134,7 +139,8 @@ class GebTestManager {
     }
 
     private Browser createBrowser() {
-        browserCreator ? browserCreator.get() : new Browser()
+        (browserCreator ? browserCreator.get() : new Browser())
```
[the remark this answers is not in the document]
Vampire: Sure, however you like it :-)

erdi:
> Btw. there is still a flak test in the AppVeyor build. The build failed and I just retriggered it to become green.

I kept emails about all of the flaky failures you experienced on your PRs and will investigate at some point to try to make it more stable and more contributor friendly.

> I actually wonder how the content cannot be found after the at-check was successfully finding the content.

That's indeed odd but I remember having a similar failure locally as well. Will investigate.

module/geb-core/src/main/groovy/geb/test/GebTestManager.groovy
Vampire: Actually currently there is another case.
If you use `browser` from the initializer of a shared field, that's before the `beforeTestClass` was called which happens in a specification interceptor and thus a browser will already be created.

PR #183, opened 2024-09-30
Remove end of maintenance notice

jonnybot0:
Work is underway to move Geb under Groovy's Apache Software Foundation umbrella. There's no need for an end-of-maintenance notice, as we will be continuing maintenance going forward.


rainboyan:
Good news for Groovy developers and the community!

Jonathan-Zollinger:
This is wonderful, wonderful news.

PR #185, opened 2024-10-06
Initial toolchain upgrade

Poundex:
Here's a start on getting the build and dependencies updated. I've updated Gradle to 8, and it will now build with Java 11. 

The full build of the Geb components, all the test suites, and the docsites & manuals are all building & passing for me locally. Most of it is also working in Geb's CI container image, except for a few of the documentation-based tasks. I don't know if they would have actually failed in CI or it is an artefact of my local build environment. That/those images will need updating with a newer JDK, anyway. 

Did a _very_ brief smoke test with the build libraries, and it seems to be working. ~~However, I did have to use a very specific version of the Selenium driver & support library (and it's not the version that's documented in the Quickstart).~~ (Turns out I had a mix of Selenium library versions on the go. 4.25.0 is fine)


doc/manual-snippets/manual-snippets.gradle
```diff
@@ -1,9 +1,23 @@
+import com.github.erdi.gradle.webdriver.WebDriverBinariesPlugin
+
+buildscript {
+    repositories {
+        maven {
+            url 'https://gitlab.com/api/v4/projects/62281009/packages/maven'
+        }
+    }
+    dependencies {
+        classpath 'com.github.erdi:webdriver-binaries-gradle-plugin:3.2.0-gradle8'
```
Poundex, line 10: I forked this so I could update a dependency (Grolifant) and quickly publish it somewhere. Might be worth Geb adopting this repo

jonnybot0: I'd actually prefer that we get rid of this plugin entirely. The functionality provided by https://github.com/erdi/webdriver-binaries-gradle-plugin can more or less be replaced by the [Selenium manager](https://www.selenium.dev/blog/2022/introducing-selenium-manager/).

Can the Grolifant update easily be separated out for the other changes? I don't want to send you down the path of updating Selenium right this moment because it'll involve updates to a lot of the test code. However, if we can skip updating Grolifant & this plugin for now without losing much of the other benefits, let's do that. Then we can focus on removing our dependency on the webdriver-binaries-gradle-plugin and let Selenium Manager do all that heavy lifting.

jonnybot0:
Thanks! This looks on the right track. 

We will need to appropriately update the CircleCI configuration for this to work. These errors:
https://app.circleci.com/pipelines/github/geb/geb/443/workflows/aba0a161-2cdd-4292-b267-5a1a754c4406/jobs/6303

Seem to indicate that Gradle isn't detecting the version of Java we need as installed in the docker image it's running on.

.circleci/config.yml seems to indicate that we're using a custom image. That image is built from https://github.com/geb/ci-docker-image. I'd bet if we added `openjdk-11-jdk` to the list of software installed in https://github.com/geb/ci-docker-image/blob/master/Dockerfile#L3, then updated the image used in the .circleci/config.yml file, this would pass.

Let's hold off on merging this until we've got the build steps passing. I'll be happy to merge as many PRs as we need on the CI image. 

While you're in there, it would be worth it to try and update the Ubuntu image it's based on, since I've already done that for the build tasks that build directly on Ubuntu.

doc/manual-snippets/manual-snippets.gradle
Poundex: The Web Drivers are now baked directly into the CI image so this plugin has been removed completely

Poundex:
Web Driver binaries plugin removed as the Web Drivers are now baked directly into the image (this was necessary for the ChromeDriver anyway due to version incompatibilities with the browser).  

Updated the CI image and the CI config and now builds are passing.

Moved to draft pending getting a new version of the CI image built and pushed.

.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2
 jobs:
   check:
     docker:
-      - image: gebish/ci:v12
```
jonnybot0, line 5: v13 with your changes is now published to Dockerhub, so you can switch to using that.

.circleci/config.yml
```diff
@@ -175,7 +170,7 @@ jobs:
 #            - .gradle-home
   non-ie:
     docker:
-      - image: gebish/ci:v7
```
jonnybot0, line 178: Same thing here, we can update to v13.

.circleci/config.yml
```diff
@@ -42,9 +42,13 @@ jobs:
       - run:
           command: |
             export WORKING_DIRECTORY=`pwd`
-            docker run -v ${WORKING_DIRECTORY}:${WORKING_DIRECTORY} -v /var/run/docker.sock:/var/run/docker.sock -w ${WORKING_DIRECTORY} -e OCI_WRITER_KEY_PASSPHRASE=$OCI_WRITER_KEY_PASSPHRASE gebish/ci:v7 /bin/bash -c "Xvfb :99 -screen 1 1280x1024x16 -nolisten tcp -fbdir /var/run > /dev/null 2>&1 & export DISPLAY=:99 ; GRADLE_OPTS=\"-Xmx1024m -XX:MaxMetaspaceSize=256m\" GRADLE_USER_HOME=\".gradle-home\" ./gradlew --no-daemon --max-workers 4 --parallel -Pci :doc:manual-snippets:check :doc:manual-snippets:fixtures:check :doc:manual-snippets:real-browser:check :doc:manual:build :doc:asciidoctor-extension:check"
-      - run:
+            docker network create dind
+            docker run -d --network dind -e DOCKER_TLS_CERTDIR="" --privileged --name dind  docker:dind
+            sleep 20
```
jonnybot0, line 47: 20 seconds seems like a long time. Does it need to be that long?

gradle/libs.versions.toml
```diff
@@ -1,12 +1,12 @@
 [versions]
-groovy = "4.0.5"
+groovy = "4.0.23"
 jetty = "9.4.44.v20210927"
 junit4 = "4.13.2"
 junit5 = "5.8.1"
-ratpack = "1.9.0"
+ratpack = "1.10.0-milestone-32"
```
jonnybot0, line 6: Will breath a little easier once @sdelamo has things moved over to GitHub pages and we can remove the Ratpack dependency. 😅

gradle/libs.versions.toml
```diff
@@ -1,12 +1,12 @@
...
 sauceConnect = "1.153"
 selenium = "4.2.2"
-spock = "2.3-groovy-4.0"
+spock = "2.4-M4-groovy-4.0"
```
jonnybot0, line 9: I'd prefer we stick to stable releases when merging to master. Milestones are fine for testing. Is there a reason you need 2.4-M4 over 2.3?

.circleci/config.yml
Poundex: The delay is waiting for the dind container to start, which includes the container itself waiting for about 15 seconds, so 20 seconds is just to make sure it has had time to fully start up

integration/geb-gradle/geb-gradle.gradle
```diff
@@ -38,10 +38,9 @@ dependencies {
     }
 }
 
-pluginBundle {
+gradlePlugin {
     website = "https://gebish.org/manual/current/#gradle-plugins"
     vcsUrl = "https://github.com/geb/geb"
-    tags = ["geb", "testing", "cloud", "browser"]
```
jonnybot0, line 44: We should carry over these tags. Just use
`      tags.set(['search', 'tags', 'for', 'your', 'hello', 'plugin'])`
rather than `tags = ...`

jonnybot0: Ah, nevermind, I see you've moved them to `plugin-definitions.gradle`. That's fine.

internal/test-support/src/main/groovy/geb/test/RemoteWebDriverWithExpectations.groovy
```diff
@@ -32,7 +32,10 @@ class RemoteWebDriverWithExpectations extends RemoteWebDriver {
     RemoteWebDriverWithExpectations(
         URL remoteAddress, List<String> ignoredCommands = DEFAULT_IGNORED_COMMANDS
     ) {
-        super(remoteAddress, new ChromeOptions().addArguments('headless'))
+        super(remoteAddress, new ChromeOptions()
+                .addArguments('headless')
+                .addArguments('--remote-allow-origins=*')
```
jonnybot0, line 37: I know it's redundant, but could we add the same TODO here as above?

module/geb-core/src/test/groovy/geb/module/DateTimeLocalInputSpec.groovy
```diff
@@ -18,11 +18,13 @@ package geb.module
 import geb.test.GebSpecWithCallbackServer
 import geb.test.browsers.Chrome
 import geb.test.browsers.RequiresRealBrowser
+import spock.lang.Ignore
 
 import java.time.LocalDateTime
 
 @Chrome
 @RequiresRealBrowser // maybe due to https://sourceforge.net/p/htmlunit/bugs/1923/
```
jonnybot0, line 26: Ignoring these tests has me a wee bit nervous that we've broken something. For a major release with a proper deprecation announcement, I'm okay with that, but that's not what we're aiming for here. I'm going to play around with these locally and have a think,.

Poundex: I've captured the tests I disabled on #188.  I've looked into two out of three of them. One of them is a change in Chrome itself. I've already fixed that one here https://github.com/Poundex/geb/tree/topic/fix-datetime-local-chrome it's just based off the upgraded branch so was waiting for that to be in before opening the MR. 

The reporting one I've investigated, and it also seems genuinely broken, but it's to do with threads so sometimes needs to be ran a few times before it can be seen failing (Spock's @RepeatUntilFailure is handy for this). I don't know if this was passing by chance before, or something has changed in the way either Gradle or Spock runs stuff in parallel. I haven't looked into fixing this one yet as it seems like quite a complicated task, but it does appear that Geb's reporting is not thread-safe. 

I haven't had a chance to look into the third one yet.

Poundex: I couldn't reproduce the issue with PageOrientedSpec so I've enabled it and it seems to be passing in CI

jonnybot0: Thanks! Glad to hear the `PageOrientedSpec` test is back online.

The other two problems (the date spec and the parallel execution spec) have me more nervous. Suppose someone is passing an ISO8601 formatted date time to Geb. Before this update, that works. After this update, it doesn't. Same problem if they're relying on parallel execution.

In this case, working around it by using JavaScript seems like hiding a breaking change. Here, the test is a set of executable documentation telling us what the `value` API is expected to do.

I haven't had a chance to really dig into this yet, but I really think we shouldn't rely on workarounds if we can avoid it at all. I think we're on the right path, but ideally the test API would continue to behave the same as before. Tooling updates shouldn't break that.

Poundex: I've done a bit more digging into the issues around the DateTimeLocalInputSpec and left a comment on #188. In short, the root cause is the Java update changes the default precision of `LocalDateTime.now()`, and as the String value of this is the only thing used as the test value then this test was always passing by accident. Whether it be caused by a Java update past 9 which changes the default precision, or be it by a datetime String coming from some other source that has a precision greater than millisecond then there is no combination of Chrome and Geb version where that test would have passed. There should definitely be some discussion on how to approach the problem, but I think that should be a focused conversation in its own thread, and not here. 

It is a similar story with the ParallelExecutionSpec, nothing is behaving any differently then it was before these changes and the issue is reproducible in current master. If you don't want to make any releases before these issues are addressed then that's fine, but I think they are outside the scope of this MR.

jonnybot0: That all makes sense. Let's merge for now and carry on discussion over on the issue.

gradle/libs.versions.toml
jonnybot0: This is the one bit of feedback I'd prefer to see addressed. Let's use 2.3, and we should be good to go.

gradle/libs.versions.toml
Poundex: Yeah, this is all we can do for now. 1.9.0 does not work with modern Gradle, and the 2.0.0 RC that is available doesn't work with our RatPack stuff.

integration/geb-gradle/geb-gradle.gradle
Poundex: Yeah this is just a change to the Gradle plugin plugin that has had to be followed

jonnybot0 (approved):

PR #190, opened 2024-10-14
Prefactor

paulk-asert:


geb.gradle
```diff
@@ -15,6 +16,15 @@ dependencies {
     site project(path: 'doc:site', configuration: 'distribution')
 }
 
+import com.github.jk1.license.filter.LicenseBundleNormalizer
```
jonnybot0, line 19: Small nitpick: can this import go at the top of the file? The linter should probably catch this, but I suppose at the moment it doesn't.

paulk-asert: Groovy allows imports anywhere but agree the usual style is to keep them at the top - moved.

PR #195, opened 2024-10-24
Update Java for Heroku

Poundex:
This change sets the JDK version for running the Heroku build, as it still uses 8 by default and we are now upgraded to 11. According to the error output and the Heroku docs this file with the single property is all that is needed, but I am not able to test it as publishing to Heroku only happens on builds of the master branch and requires secrets to be injected.

### DIFF
--- a/system.properties
+++ b/system.properties
@@ -1,0 +1,1 @@
+java.runtime.version=11


jonnybot0:
I'm okay with using 11 in the pipeline. Actually, I'd be okay going all the way to 21, but can we set the compile target to stay at 1.8? 
```
sourceCompatibility = JavaVersion.VERSION_1_8
targetCompatibility = JavaVersion.VERSION_1_8
```

We can make a decision about when we want to have a major release that forces users to upgrade Java based on end of security support from major JDK publishers. 

* https://endoflife.date/eclipse-temurin
* https://www.oracle.com/java/technologies/java-se-support-roadmap.html

jonnybot0 (approved):

Issue #194, opened 2024-10-23
The DateTimeLocalInput module will pass values that will be rejected by Chrome


Poundex:
Chrome is very strict when directly setting the value of a date/time input. It will accept the format `YYYY-MM-dd'T'HH:mm[:ss[.SSS]]` and nothing else. A date/time string with any more precision than this will not be truncated, but vetoed with an error that will be logged by the console:
![chrome_datetime](https://github.com/user-attachments/assets/545df0d6-a349-4c50-a6c0-fe5591f7eadc)
but not communicated to the driver in any way, meaning that from the test's point of view a requested change has simply not happened. 

This module exposes two methods for setting the value of the input, one that takes a String and one that takes a LocalDateTime. The one that takes a String passes the value straight through without any validation or processing. This will succeed if it matches Chrome's expected format, and fail (silently) if it doesn't - even if it is a perfectly valid ISO8601 date/time string.
The method that takes a LocalDateTime passes the toString value. If the LocalDateTime instance was constructed with `LocalDateTime.now()` then this will behave differently after Java 9. Before Java 9 `LocalDateTime.now()` had millisecond precision, and so a `toString()` call would match Chrome's expected format. But after Java 9 it has nanosecond precision, and so the extra digits in the fraction-of-second part would cause the set value to be rejected by Chrome.

jonnybot0:
This was fixed by df7d709, so I think we can safely close it.

PR #196, opened 2024-10-24
#194 - Reformat any date time values passed to the DateTimeLocalInput module into a format that will be accepted by Chrome

Poundex:
This change will reformat any date time values passed to the module into a format that will work in the browser. 

If the provided value has more precision than would be accepted, then it is truncated and a warning is emitted.

### DIFF
--- a/module/geb-core/src/main/groovy/geb/module/DateTimeLocalInput.groovy
+++ b/module/geb-core/src/main/groovy/geb/module/DateTimeLocalInput.groovy
@@ -15,18 +15,36 @@
  */
 package geb.module
 
-import java.time.LocalDateTime
+import groovy.util.logging.Slf4j
 
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
+import java.time.format.DateTimeFormatterBuilder
+
+import static java.time.temporal.ChronoField.*
+
+@Slf4j
 class DateTimeLocalInput extends AbstractInput {
 
+    private static final DateTimeFormatter DATE_TIME_FORMAT = new DateTimeFormatterBuilder()
+            .append(DateTimeFormatter.ISO_LOCAL_DATE)
+            .appendLiteral('T')
+            .appendValue(HOUR_OF_DAY, 2)
+            .appendLiteral(':')
+            .appendValue(MINUTE_OF_HOUR, 2)
+            .appendLiteral(':')
+            .appendValue(SECOND_OF_MINUTE, 2)
+            .appendFraction(MILLI_OF_SECOND, 0, 3, true)
+            .toFormatter()
+    
     final String inputType = 'datetime-local'
 
     void setDateTime(LocalDateTime dateTime) {
-        value(dateTime.toString())
+        value(reformat(dateTime))
     }
 
     void setDateTime(String iso8601FormattedDateTime) {
-        value(iso8601FormattedDateTime)
+        setDateTime(LocalDateTime.parse(iso8601FormattedDateTime))
     }
 
     LocalDateTime getDateTime() {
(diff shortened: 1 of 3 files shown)


module/geb-core/src/main/groovy/geb/module/DateTimeLocalInput.groovy
```diff
@@ -39,4 +57,13 @@ class DateTimeLocalInput extends AbstractInput {
         super.isTypeValid(type) || type == "text"
     }
 
+    private static String reformat(LocalDateTime localDateTime) {
+        String inputValue = localDateTime.format(DATE_TIME_FORMAT)
+        if(localDateTime.toString() != inputValue) {
+            log.warn("The datetime value {} was truncated to {} as it was being used to set the value of a <input type=\"datetime-local\" />",
```
jonnybot0, line 63: I'm not sure this warning is necessary. If the user gave us a LocalDateTime, they're basically trusting Geb to format it for them. This warning is guaranteed to show up on any version of Java later than 9, since the default `toString` will change.

I think the better thing to do would be provide an additional method that lets people supply their own `DateTimeFormatter` if they want. There may be use cases where that's relevant that we haven't foreseen. Even if we don't do that, I don't see a compelling reason for this warning. After all, "if you're always on red alert, you never are."

Better to just put the formatter in by default and give people an escape hatch if it doesn't meet their needs.

module/geb-core/src/test/groovy/geb/module/DateTimeLocalInputSpec.groovy
```diff
@@ -72,10 +71,14 @@ class DateTimeLocalInputSpec extends GebSpecWithCallbackServer {
...
 
         then:
-        input.dateTime == dateTime.plusDays(1)
+        input.dateTime == truncated(dateTime.plusDays(1))
 
         where:
         dateTime = LocalDateTime.now()
     }
+    
+    private static LocalDateTime truncated(LocalDateTime ldt) {
```
jonnybot0, line 80: If we do the formatting in a standard way in `DateTimeLocalInput`, is this truncation even necessary? Wouldn't it get handled automatically in the `reformat` method?

module/geb-core/src/main/groovy/geb/module/DateTimeLocalInput.groovy
```diff
@@ -39,4 +57,13 @@ class DateTimeLocalInput extends AbstractInput {
         super.isTypeValid(type) || type == "text"
     }
 
+    private static String reformat(LocalDateTime localDateTime) {
```
jonnybot0, line 60: 1. Can we just leave this method public? Seems like someone might reasonably want to capture the formatted date value as part of their test.
2. How about `format`, rather than `reformat`?

drop the re just format
;P

module/geb-core/src/test/groovy/geb/module/DateTimeLocalInputSpec.groovy
Poundex: Yes, this is to truncate the expected values in the tests so they match the properly truncated actuals

module/geb-core/src/main/groovy/geb/module/DateTimeLocalInput.groovy
```diff
@@ -15,18 +15,44 @@
...
 import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
+import java.time.format.DateTimeFormatterBuilder
+
+import static java.time.temporal.ChronoField.*
 
+@Slf4j
 class DateTimeLocalInput extends AbstractInput {
 
+    /* codenarc-disable */
```
jonnybot0, line 29: Final nit: let's disable the specific rules we need to disable:

```
    // codenarc-disable DuplicateNumberLiteral
    // codenarc-disable DuplicateStringLiteral
```

Do that, and you're good to merge.

module/geb-core/src/main/groovy/geb/module/DateTimeLocalInput.groovy
```diff
@@ -15,18 +15,44 @@
...
+    private static final DateTimeFormatter DATE_TIME_FORMAT = new DateTimeFormatterBuilder()
```
jonnybot0, line 30: I was generally surprised that I could not find a pre-built formatter that matched this format, but the source of `java.time.format.DateTimeFormatter` certainly didn't have any.

module/geb-core/src/main/groovy/geb/module/DateTimeLocalInput.groovy
jonnybot0: Actually, I'll just do this as a follow-up. It's merging time!

PR #203, opened 2024-11-14
update documentation

paulk-asert:


README.md
```diff
@@ -12,5 +12,9 @@ Please see [CONTRIBUTING.md](https://github.com/geb/geb/blob/master/CONTRIBUTING
 
 ## Submitting issues
 
-If you'd like to submit an issue against Geb then please use [the issue tracker for geb/issues GitHub project](https://github.com/geb/issues/issues).
-Please avoid submitting usage questions as issues and use [the user mailing list](https://groups.google.com/forum/#!forum/geb-user) instead.
\ No newline at end of file
+If you'd like to submit an issue against Geb then please use the [Geb GitHub issue tracker](https://github.com/apache/groovy-geb/issues).
```
jonnybot0, line 15: This section does need changed, but I think we need to build some consensus. @Poundex and @sdelamo have both expressed a preference for using GitHub discussions. I know that the mailing list is mandatory for Apache projects, but can we point folks to https://github.com/apache/groovy-geb/discussions first, out of deference to the contributors?

I think a better wording would be:

```
If you'd like to submit an issue against Geb then please use the [GitHub Discussions](https://github.com/apache/groovy-geb/discussions) first. Those are better suited to the back-and-forth required to clarify an issue into something actionable.

+ Please avoid submitting usage questions as issues and instead join and ask your question on the `geb-users@groovy.apache.org` mailing list:
+ [Browse](https://lists.apache.org/list.html?groovy-users@groovy.apache.org)
+ [Subscribe](geb-users-subscribe@groovy.apache.org)
+ [Unsubscribe](geb-users-unsubscribe@groovy.apache.org)
```

In short, if your question is, "How do I...", mailing list. If your problem is, "This doesn't work..." go to the discussions and we'll work with you to write an issue.

Would that work?

Issue #226, opened 2024-12-19
Move maven coordinates to be under org.apache.groovy.geb rather than org.gebish


paulk-asert:
This will be for the geb jars. We need to consider a similar move for the gradle plugins. We can do that as a separate issue.

I bumped the version from 7.1-SNAPSHOT to 8.0-SNAPSHOT since I think maven coordinate change warrants that but the team needs to agree on that.

paulk-asert:
Currently there's a snapshot published here:
https://repository.apache.org/content/repositories/snapshots/org/apache/groovy/geb

paulk-asert:
Proposed PR merged. I needed that to test the Jenkins build but we can rollback if anyone sees issues.

PR #228, opened 2024-12-19
Move maven coordinates to be under org.apache.groovy.geb rather than …

paulk-asert:
…org.gebish #226


paulk-asert:
The tests are currently failing. It is trying to check:
  https://oss.sonatype.org/content/repositories/snapshots/org/apache/groovy/geb/
We haven't published any snapshots under the new coords, so even if we fixed the URL it would still break at this stage.
I haven't worked out yet whether we need to cater for both old and new URL prefixes.

Issue #237, opened 2024-12-23
Bump org.asciidoctor.jvm.convert version to  4.0.4


paulk-asert:

paulk-asert:
Merged.

PR #239, opened 2024-12-23
Bump org.asciidoctor.jvm.convert version to 4.0.4 #237

paulk-asert:


jonnybot0 (approved):
Seems like a quick & harmless change.

PR #240, opened 2024-12-23
Fix codenarc space error

jonnybot0:
I noticed that codenarc was failing in CircleCI due to SpaceAfterCommentDelimiter violations. This was a quick fix with a regex, so I'm going to just merge it.


paulk-asert:
We should check the Codenarc config between what the GitHub action is running and what CircleCI is running. If CircleCI is failing, I would have thought the GitHub action should have failed too.

jonnybot0:
The main difference was that the previous style check workflow was only running codenarcMain. `check` will also run `codenarcTest`, along with `test`.

PR #241, opened 2024-12-23
Make test step add codenarc and doc site build

jonnybot0:
This makes the base test step cover the same ground as the CircleCI build by adding the `ci` property and using `check`, not just `test`. The main difference is that this will run CodeNarc and copy reports from submodules into the root directory.

It seemed reasonable to me to run these all in the same step.


paulk-asert:
You should be able to delete style-check.yml. At the start, build and test was failing, so I just wanted the codenarc checks to be green - but fine to run them all together.

Issue #245, opened 2025-01-15
Build a src distribution for releasing


paulk-asert:
All ASF projects need to provide source releases.

The proposed PR builds a Zip source release. ASF releases must in general not contain binary files and at present the wrapper is considered a non-allowable binary file, so it is excluded. The CONTRIBUTING.md file is updated accordingly with steps to bootstrap the wrapper files if using a source distribution. The details follow what Groovy is doing but there are other possible options if folks aren't happy with the approach shown here.

Only a small percentage of folks typically use the source distribution, so I think the approach is a reasonable compromise.

PR #246, opened 2025-01-15
Build a src distribution for releasing #245

paulk-asert:
Builds a  source zip suitable for releasing.

ASF release policy requires source releases have no binary files, so the gradle wrapper is excluded. It can easily be added back in using the `wrapper` task or the `bootstrap` project.


paulk-asert:
The signing part will probably need some additional work.

paulk-asert:
Proposed PR merged.

PR #254, opened 2025-01-22
prepare for gathering apache manuals from changed maven coordinates

paulk-asert:


paulk-asert:
We may want to also add the staging URL in the repositories list:

    maven { url = 'https://repository.apache.org/content/repositories/staging/' }

This would allow publishing the site prior to finalising a new release. We need to finalise our release process first before we know exactly what is needed.

PR #260, opened 2025-02-12
Replace google analytics with matomo in old manuals

sdelamo:
It also replaces cloudflare links from fontawesome with a local copy of the resources.


paulk-asert (approved):
LGTM. We could re-use some js/css/font files from the main Groovy website but having separate Geb copies seems safer to me, so happy with the PR as is.

PR #263, opened 2025-02-21
Add a few more GitHub actions to replace some CircleCI

jonnybot0:
Following the [discussion with @paulk-asert](https://github.com/apache/groovy-geb/issues/213#issuecomment-2675418754), I started going through the remaining actions on CircleCI to see about moving them over to GitHub Actions or Jenkins.

This removes the easy-to-reach ones. The remaining few will require some secrets management, as they require API keys to BrowserStack, SauceLabs, or LambdaTest.

I've got access to those via the Geb org that Marcin shared with me when handing things over.

Maintaining [Geb's integration with those providers](https://gebish.org/manual/current/#cloud-browser-testing) will require setting those up as shared secrets. I don't appear to have access to do that in GitHub like I did for the geb organization (https://github.com/organizations/geb/settings/secrets/actions).

I could probably get that done with a request to Apache Infra, for either GH Actions or Jenkins. If we do that, we can completely remove the CircleCI build.

I went ahead and used the build cache on the new files. I'm trusting that the Gradle tasks are deterministic enough that we needn't re-run them (see, e.g. https://blog.gradle.org/stop-rerunning-tests).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,34 +15,6 @@
 
 version: 2
 jobs:
-  check:
-    docker:
-      - image: eclipse-temurin:21-jdk
-    resource_class: large
-    environment:
-      GRADLE_OPTS: -Xmx1024m -XX:MaxMetaspaceSize=256m
-      GRADLE_USER_HOME: .gradle-home
-    steps:
-      - checkout
-      - restore_cache:
-          keys:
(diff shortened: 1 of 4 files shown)


jonnybot0:
(Going with commit-then-review, but @me if we need to clean this up somehow)

PR #267, opened 2025-04-24
Fix a race-condition in reporting with parallel execution

### LINKED ISSUE #188: Investigate unhealthy tests
- **ParallelExecutionSpec** - _Flaky in CI_
```
ParallelExecutionSpec > GebReportingSpec supports parallel execution at feature level FAILED
    Condition not satisfied:

    reportFileTestCounterPrefixes("SpecRunningIterationsInParallel1") == (["000"] * 2) + (1..4)*.toString()*.padLeft(3, "0")
    |                                                                 |           |    |         |           |
    [000, 000, 001, 002, 003, 004, 004]                               false       |    |         |           [001, 002, 003, 004]
                                                                                  |    |         [1, 2, 3, 4]
                                                                                  |    [000, 000, 001, 002, 003, 004]
                                                                                  [000, 000]
        at geb.spock.ParallelExecutionSpec.GebReportingSpec supports parallel execution at feature level(ParallelExecutionSpec.groovy:186)

```

- **PageOrientedSpec** - _Flaky in CI_
```
PageOrientedSpec > verify the Page API works for '#contentName' content > verify the Page API works for 'linkUsingPageInstance' content FAILED
    org.openqa.selenium.StaleElementReferenceException: The element seems to be disconnected from the DOM.  This means that a user cannot interact with it.
    For documentation on this error, please visit: https://selenium.dev/exceptions/#stale_element_reference
    Build info: version: '4.2.2', revision: '683ccb65d6'
    System info: host: '075353d81e19', ip: '192.168.176.3', os.name: 'Linux', os.arch: 'amd64', os.version: '5.15.0-1057-aws', java.version: '11.0.24'
    Driver info: driver.version: IDataDriver
        at app//org.openqa.selenium.htmlunit.HtmlUnitDriver.assertElementNotStale(HtmlUnitDriver.java:899)
        at app//org.openqa.selenium.htmlunit.HtmlUnitWebElement.assertElementNotStale(HtmlUnitWebElement.java:591)
        at app//org.openqa.selenium.htmlunit.HtmlUnitWebElement.verifyCanInteractWithElement(HtmlUnitWebElement.java:249)
        at app//org.openqa.selenium.htmlunit.HtmlUnitWebElement.click(HtmlUnitWebElement.java:131)
        at app//geb.navigator.DefaultNavigator.click(DefaultNavigator.groovy:672)
        at app//geb.navigator.DefaultNavigator.click(DefaultNavigator.groovy:687)
        at app//geb.content.TemplateDerivedPageContent.click(TemplateDerivedPageContent.groovy:83)
        at geb.PageOrientedSpec.verify the Page API works for '#contentName' content(PageOrientedSpec.groovy:73)
```

- **DateInputSpec** - _Functionality broken in recent ChromeDriver_
It seems this test is failing due to broken functionality in recent versions of ChromeDriver, where setting a value of an `<input type="datetime-local" />` does not work as expected. There is some discussion of the issue on [this php-webdriver thread](https://github.com/php-webdriver/php-webdriver/issues/709). There is an [existing ChromeDriver ticket](https://issues.chromium.org/issues/42322342) but it doesn't look like it's getting fixed. A common workaround seems to be using JavaScript to set the value, rather than manipulating directly, [like in this example](https://github.com/teamcapybara/capybara/blob/e8874a1cbc3e967b74621fb243c0ae514625900f/lib/capybara/selenium/node.rb#L255).

Vampire:
Without this PR the parallel execution tests failed one to many times
when executing them 10 times in a row due to wrong report group being used.

With this PR this race condition is fixed and it did not fail in the 500 runs
I made to verify, as I don't see how to write a reproducible test case for this situation.

Partly fixes #188

### DIFF
--- a/module/geb-core/src/main/groovy/geb/test/GebTestManager.groovy
+++ b/module/geb-core/src/main/groovy/geb/test/GebTestManager.groovy
@@ -97,6 +97,11 @@ class GebTestManager {
         this.testClass.get().push(testClass)
         currentTestName.set(testName)
         if (reportingEnabled) {
+            // it happens that beforeTestClass for test class A is run on a thread
+            // and next beforeTest (and thus test execution) for tests from class B are run on that thread
+            // as the browser is thread-local reused, we need to set the report group here or the report
+            // eventually is written to the wrong report group which is confusing and makes some tests flaky
+            getBrowser().reportGroup(testClass)
             testCounter.get().push(nextTestCounter(testClass))
             perTestReportCounter.get().push(1)
         }


module/geb-core/src/main/groovy/geb/test/GebTestManager.groovy
```diff
@@ -97,6 +97,11 @@ class GebTestManager {
         this.testClass.get().push(testClass)
         currentTestName.set(testName)
         if (reportingEnabled) {
+            // it happens that beforeTestClass for test class A is run on a thread
+            // and next beforeTest (and thus test execution) for tests from class B are run on that thread
+            // as the browser is thread-local reused, we need to set the report group here or the report
+            // eventually is written to the wrong report group which is confusing and makes some tests flaky
+            getBrowser().reportGroup(testClass)
```
jonnybot0, line 104: This seems a sensible default. Should still be overrideable from tests. 👍

jonnybot0 (approved):
Approved, assuming tests pass.

module/geb-core/src/main/groovy/geb/test/GebTestManager.groovy
Vampire: It's the same as it should always have been and also was if not by chance a thread executed the `beforeTestClass` of one class and then immediately the `beforeTest` of a different class. And yes, this of course is still overridable. :-)

PR #270, opened 2025-05-07
Fix the driver caching so that global and per-thread cache do not step on each others toes

Vampire:


Vampire:
Partly fixes #188

Vampire:
Without this fix, the problem reproduced within the first 10 tries to run the parallel tests.
With this fix the first 1000 tries went through without problem and without hang.

PR #271, opened 2025-05-14
Separate the dynamic page content of PageOrientedSpec to only the page that expects it

Vampire:
With JavaScript enabled, the JavaScript thread modifies the page and with unlucky timing
the link element needed is either not present when checked for in at-checking, or is stale
when trying to click it. This makes the tests not actually needing the dynamic element
flaky, failing occasionally for content not found or stale elements.

### DIFF
--- a/module/geb-core/src/test/groovy/geb/PageOrientedDynamicSpec.groovy
+++ b/module/geb-core/src/test/groovy/geb/PageOrientedDynamicSpec.groovy
@@ -1,0 +1,116 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package geb
+
+import geb.test.GebSpecWithCallbackServer
+import spock.lang.Unroll
+
+class PageOrientedDynamicSpec extends GebSpecWithCallbackServer {
+
+    def setupSpec() {
+        callbackServer.get = { req, res ->
+            def path = req.requestURI == "/b" ? "b" : "a"
+            def other = path == "b" ? "a" : "b"
+            res.outputStream << "<html>"
+            if (path == "b") {
+                res.outputStream << """
+                <head>
+                    <script type="text/javascript" charset="utf-8">
+                    setTimeout(function() {
+                        document.body.innerHTML += '<div id="dynamic">dynamic</div>';
+                    }, 100);
+                    </script>
+                </head>"""
+            }
+            res.outputStream << """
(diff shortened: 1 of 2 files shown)


Vampire:
Fixes another part of #188

jonnybot0:
The more I look at it, the more unsure I am how unlucky timing could be a factor here. The JS on the page just appends to the body. It doesn't manipulate the existing link, so I don't think it should cause a stale element exception as described in #188.

I've got half a mind to say we close #188 as "2/3 are good enough" and revisit this if we see further flakiness down the line. If that's not good enough, I'd say just split those tests out into their own spec.

jonnybot0:
Hm. Okay, I can sorta reproduce the issue by reducing the timeout in the JS (see this patch). Only running it for the content that failed in #188, but it works equally well for the `link`. Fails on about 11/100 runs.

```
diff --git a/module/geb-core/src/test/groovy/geb/PageOrientedSpec.groovy b/module/geb-core/src/test/groovy/geb/PageOrientedSpec.groovy
index d1d164a8..b738b12e 100644
--- a/module/geb-core/src/test/groovy/geb/PageOrientedSpec.groovy
+++ b/module/geb-core/src/test/groovy/geb/PageOrientedSpec.groovy
@@ -36,7 +36,7 @@ class PageOrientedSpec extends GebSpecWithCallbackServer {
                     <script type="text/javascript" charset="utf-8">
                     setTimeout(function() {
                         document.body.innerHTML += '<div id="dynamic">dynamic</div>';
-                    }, 100);
+                    }, 1);
                     </script>
                 </head>
                 <body>
@@ -49,7 +49,6 @@ class PageOrientedSpec extends GebSpecWithCallbackServer {
     }
 
     def setup() {
-        driver.webClient.options.javaScriptEnabled = false
     }
 
     def "verify our server is configured correctly"() {
@@ -90,7 +89,7 @@ class PageOrientedSpec extends GebSpecWithCallbackServer {
         at PageOrientedSpecPageA
 
         where:
-        contentName << ["link", "linkUsingPageInstance"]
+        [contentName, index] << [["linkUsingPageInstance"], (1..100).toList()].combinations()
     }
 
     @Unroll
```

Having the JS that's at the root of the problem and the tests that require it in a separate spec still seems the better path.

Vampire:
> particularly to squash some flakiness that we haven't seen much of lately. I wonder if it was peculiar to the old CircleCI stuff that was more of a going concern back in October.

I got the flaky behavior consistently within the first 10 tries.
Maybe on GHA the timing is just slightly different than on CCI or locally and thus it does not happen that often,
but it definitely still is a problem.

> The more I look at it, the more unsure I am how unlucky timing could be a factor here. 

That was not an assumption, but a fact.
Derived from quite some logging breakpoints and system out debugging.
The JS might just append, but it HtmlUnit then rebuilds the model completely which then causes the element to not be found or being stale, depending on the concrete timing.

>  revisit this if we see further flakiness down the line.

Which is ... now.
This still **is** a problem.

> If that's not good enough, I'd say just split those tests out into their own spec.

I will

Vampire:
Indeed the fix here was not even sufficient.
It solved the not-found and stale for the non-dynamic features.
But now after trying to run the extracted spec for the dynamic features, the not-found does not happen but stale still happen,
as `to PageOrientedDynamicSpecPageA` for those still shows that problem.
Good that we talked about it. :-D

Vampire:
I hopefully fixed it, let's see what the next 10_000 runs say. :-)

Vampire:
First 1_000 tries were successful, so update is sent :-)

jonnybot0 (approved):
Thanks for going the extra mile. Looks good!

jonnybot0:
Went ahead and merged since only failing tests were from the Geb site move, which were addressed elsewhere.

Vampire:
Ok, 5_396 tries were successful, then I needed to debug a different Gradle build and thus got a timeout on the Gradle journal lock.
But I'd say it looks fine. :-)

PR #273, opened 2025-05-22
Support GString as JS execution argument

Vampire:


jonnybot0 (approved):
This seems like a must-have for Groovy.

PR #275, opened 2025-05-26
Support Closure<CharSequence> JS execution script argument

Vampire:


doc/manual-snippets/src/test/groovy/javascript/ExecutingArbitraryCodeSpec.groovy
```diff
@@ -50,4 +50,25 @@ class ExecutingArbitraryCodeSpec extends GebSpecWithCallbackServer {
...
+                function someJsMethod(a,b) {
+                }
+            """)
+        }
+
+        expect:
+        // tag::closure[]
+        js.exec(1, 2) {
+            """
+                someJsMethod(1, 2);
```
jonnybot0, line 67: Should this example be passing in arguments here, rather than just repeating `1, 2`?

module/geb-core/src/main/groovy/geb/js/JavascriptInterface.groovy
```diff
@@ -53,8 +53,11 @@ class JavascriptInterface {
             jsArgs = args[0..(args.size() - 2)]
         }
 
+        if (script instanceof Closure) {
+            script = script()
+        }
         if (!(script instanceof CharSequence)) {
-            throw new IllegalArgumentException("The last argument to the js function must be string-like")
+            throw new IllegalArgumentException("The last argument to the js function must be string-like or a Closure returning a string-like")
```
jonnybot0, line 60: I think this is legitimate functionality, but it makes me wonder why we'd only accept a `Closure`. Why not other functional interfaces, like `Callable`? Gets tricky, of course, but your other PRs make me wonder if we aren't actually hankering for a better designed API than this somewhat ad-hoc "throw stuff in the method" JavaScriptInterface.

Can you say a bit more about the felt need behind this PR? What led you here?

Vampire: > "throw stuff in the method"

Well, good ol' Groovy duck-typing, isn't it? :-D

> Why not other functional interfaces?
> Can you say a bit more about the felt need behind this PR? What led you here?

Well, yeah, other FI could be supported of course.
My main intention was to support Syntax that feels more natural and readable to me.
Without this, you can for example do 
```groovy
js.exec 1, 2, '''
    someJsMethod(arguments[0], arguments[1]);
    // lots of javascript
    return true;
'''
```
But I having multiple comma-separated arguments and additionally multi-line without parens feels a bit strange to me,
besides that you first have the arguments, then the script they are for and that intermixed.

Or you can do 
```groovy
js.exec(1, 2, '''
    someJsMethod(arguments[0], arguments[1]);
    // lots of javascript
    return true;
''')
```
but that also imho does not read super nice.

So my target was to have the new syntax
```groovy
js.exec(1, 2) {
    '''
        someJsMethod(arguments[0], arguments[1]);
        // lots of javascript
        return true;
    '''
}
```
where the arguments are comma-separated in parentheses and the script comes in the following block outside the parentheses, to have a clearer separation between the script and the arguments for the script and have an imho also much better readable syntax.

PR #279, opened 2025-07-07
Prep for release

jonnybot0:
Doing some work here to prepare for the 8.0 release.

The big change here is that the `publishJarsAndManual` task had been removed from most of the modules by f2a951fb53f73700df4c3b4a613b8c67369715f. I suspect this was inadvertent. 

The result of this was that while our snapshot build was running, it wasn't actually publishing anything. See https://ci-builds.apache.org/job/Groovy/job/Geb%20snapshot%20build/lastBuild/console and the publication dates at https://repository.apache.org/content/repositories/snapshots/org/apache/groovy/geb/ for confirmation of this. I did a local test publication of the [geb-waiting module](https://repository.apache.org/content/repositories/snapshots/org/apache/groovy/geb/geb-waiting/) to verify that this change worked correctly. Compare the results of `./gradlew publishJarsAndManual --dry-run` with master for further confirmation.

I've also updated the developer list. I included @paulk-asert, @sdelamo, and myself for now, though we may wish to update this soonish. I tried to think, "Who should be contacted about this project?" as the way to clarify the choice.

I toyed around with [publishing maven package relocation information](https://docs.gradle.org/8.14.3/userguide/publishing_maven.html#publishing_maven:relocation) for the old org.gebish coordinates. I got a 403 trying to write to the org.gebish package in Apache's nexus repository. I imagine we could still build this POM and publish it to central, but we may need to go another route.

### DIFF
--- a/buildSrc/src/main/groovy/geb.publishing.gradle
+++ b/buildSrc/src/main/groovy/geb.publishing.gradle
@@ -47,14 +47,16 @@ publishing {
             }
             developers {
                 developer {
-                    id = "alkemist"
-                    name = "Luke Daley"
-                    roles = ["Founder"]
+                    id = "paulk-asert"
+                    name = "Paul King"
                 }
                 developer {
-                    id = "erdi"
-                    name = "Marcin Erdmann"
-                    roles = ["Lead"]
+                    id = "jonnybot"
+                    name = "Jonny Carter"
+                }
+                developer {
+                    id = "sdelamo"
+                    name = " Sergio del Amo"
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/groovy/geb.publishing.gradle
```diff
@@ -47,14 +47,16 @@ publishing {
...
                 developer {
-                    id = "erdi"
-                    name = "Marcin Erdmann"
-                    roles = ["Lead"]
+                    id = "jonnybot"
+                    name = "Jonny Carter"
+                }
+                developer {
+                    id = "sdelamo"
+                    name = " Sergio del Amo"
```
Vampire, line 59:
```suggestion
                    name = "Sergio del Amo"
```

sdelamo (approved):

PR #310, opened 2026-03-09
Task/geb testcontainers

cbmarcum:
Testcontainers integration for review.


jonnybot0:
A question you asked in your mail:
>  Should we change the packages from grails.plugin.geb or leave them alone?

I would say we should move the packages since the plan is to incorporate this into geb itself. This won't be a Grails plugin hosted within Geb, but a native integration with testcontainers.

integration/geb-testcontainers/src/main/groovy/geb/testcontainers/support/delegate/DownloadSupportDelegate.groovy
```diff
@@ -0,0 +1,173 @@
...
+ * <p>
+ * As method parameter names are not available in the Geb artifacts we are delegating manually to
+ * get the best possible IDE support and user experience.
+ *
+ * @author Mattias Reichel
+ * @since 4.2
+ */
+@CompileStatic
+@SelfType(ContainerGebSpec)
+trait DownloadSupportDelegate implements DownloadSupport {
```
jonnybot0: I suspect that the traits in this package may be something we could simplify and remove for the Geb library. If the Grails test containers integration wants to keep doing things this way, that's totally fine. For the Geb library, my intuition is that we can provide a simpler base class for people to build off.

I recognize, you're probably following this pattern in no small part to provide continuity with the old Grails plugin, so no complaints. This also may be something that's better shown than told. I'm going to try to come up with an alternate implementation and open a PR into your branch to demonstrate what I mean. If I can't get it working, you'll know that it was just me overthinking! 😄

jonnybot0 (changes requested):

integration/geb-testcontainers/src/main/groovy/geb/testcontainers/support/delegate/DownloadSupportDelegate.groovy
cbmarcum: Sure,  let me know if you have a chance to look into it.

cbmarcum:
The packages have been changed to geb.testcontainers consistent with other module conventions.

integration/geb-testcontainers/geb-testcontainers.gradle
```diff
@@ -0,0 +1,142 @@
...
+
+tasks.named('test') {
+    testLogging {
+        events 'passed', 'skipped', 'failed'
+    }
+
+    systemProperty('geb.container.recording.mode', 'RECORD_ALL') // also RECORD_FAILING
+}
+
+tasks.register('integrationTest', Test) {
```
jonnybot0, line 83: The way this task is wired up seems to prevent someone from running just one test. I think I've fixed that in my feedback PR at https://github.com/apache/groovy-geb/pull/313. @me if that doesn't look right, though.

integration/geb-testcontainers/src/integration-test/groovy/geb/testcontainers/PortConfigFromClassSpec.groovy
```diff
@@ -0,0 +1,58 @@
...
+ * Test spec to verify that the correct port configuration is being used.
+ * Some reference taken from:
+ * groovy-geb/module/geb-core/src/test/groovy/geb/conf/ConfigurationLoaderSpec.groovy
+ */
+@Narrative("To verify that class field overrides config file.")
+class PortConfigFromClassSpec extends ContainerGebSpec {
+
+    static TestFileServer server
+
+    int hostPort = 8000
```
jonnybot0: This field seems effectively unused. That makes me wonder what exactly this test is testing. `"should use the class field value"` doesn't seem to describe what the test is doing.

Likewise, this test has been flaky for me locally. At first I thought it was because I had other containers running on that port, but that was not it.

I've pushed what I think is a fix to my review branch, but I'd like to know what the purpose of this test is. Just because it passes doesn't matter if I don't understand what functionality this test is meant to verify.

integration/geb-testcontainers/src/main/groovy/geb/testcontainers/support/delegate/DownloadSupportDelegate.groovy
jonnybot0: Alrighty. I've got a PR up with a proposed refactor. https://github.com/apache/groovy-geb/pull/313

Let me know what you think. I'm not pretending my approach is more perfect, though it does seem a bit more consistent with the rest of Geb (extending a specific base class).

integration/geb-testcontainers/src/integration-test/groovy/geb/testcontainers/ContainerFileDetectorSpockSpec.groovy
```diff
@@ -0,0 +1,54 @@
...
+
+    def setupSpec() {
+        ServiceRegistry.setInstance(ContainerFileDetector, new UselessContainerFileDetector())
+    }
+
+    def cleanupSpec() {
+        ServiceRegistry.setInstance(ContainerFileDetector, null)
+    }
+
+    @PendingFeature(reason = 'https://github.com/apache/grails-geb/pull/146#issuecomment-2691433277')
```
jonnybot0: Nitpicky thing, but this and some other references to `grails` seem somewhat out of place. Same thing for the `grails:gebConfigUsed` properties used elsewhere. I love our friends on Grails, but I fear this might be a wee bit confusing to new contributors.

integration/geb-testcontainers/src/integration-test/groovy/geb/testcontainers/PortConfigFromClassSpec.groovy
cbmarcum: The possible settings of hostPort are described in the [readme host port section](https://github.com/apache/groovy-geb/tree/task/geb-testcontainers/integration/geb-testcontainers#host-port).
The GebConfig.groovy for this test sets it to 8090 (option 2).  This test is meant to prove that setting it with a class level field in the test (Option 3) will override the option 2 configuration.
See the WebDriverContainerHolder findServerPort method.

integration/geb-testcontainers/src/integration-test/groovy/geb/testcontainers/ContainerFileDetectorSpockSpec.groovy
cbmarcum: not nitpicky at all. I've updated the capability names to geb:* And it seems you have fixed that test's need for pending.

integration/geb-testcontainers/src/main/groovy/geb/testcontainers/support/delegate/DownloadSupportDelegate.groovy
cbmarcum: @jonnybot0 , I've merged your #313 changes and other cleanups.  Please have another go at it.

Dockerfile
jonnybot0, line 1: So, I think that building this Dockerfile is a separate concern from testcontainers integration that got folded into this PR. Can we back that out and review it on its own, separately? We made a decision a while back to use the off-the-shelf Docker containers for most build concerns, rather than publishing our own just for CI. Unless I'm misreading the PR, we don't really need it, the docker-entrypoint.sh, build-in-docker.sh, and so on.

gradle/libs.versions.toml
```diff
@@ -63,11 +65,16 @@ sauceConnect = { module = 'com.saucelabs:ci-sauce', version.ref = 'sauceConnect'
...
+selenium-firefox-driver = { module = "org.seleniumhq.selenium:selenium-firefox-driver", version.ref = "selenium" }
+selenium-edge-driver = { module = "org.seleniumhq.selenium:selenium-edge-driver", version.ref = "selenium" }
+selenium-remote-driver = { module = "org.seleniumhq.selenium:selenium-remote-driver", version.ref = "remote-driver" }
+testcontainers-selenium = { module = "org.testcontainers:selenium", version.ref = "testcontainers" }
 spock-core = { module = 'org.spockframework:spock-core', version.ref = 'spock' }
 spockCoreGroovy25 = 'org.spockframework:spock-core:1.3-groovy-2.5'
 spockCoreGroovy3 = 'org.spockframework:spock-core:2.3-groovy-3.0'
 systemLambda = 'com.github.stefanbirkner:system-lambda:1.2.1'
-testcontainers-selenium = 'org.testcontainers:selenium:1.20.4'
+# testcontainers-selenium = 'org.testcontainers:selenium:1.20.4'
```
jonnybot0: Nit: Why this commented out line?

doc/manual-snippets/real-browser/src/test/resources/GebConfig.groovy
```diff
@@ -16,7 +16,7 @@
  *  specific language governing permissions and limitations
  *  under the License.
  */
-import org.openqa.selenium.chrome.ChromeOptions
+import org.openqa.selenium.firefox.FirefoxOptions
```
jonnybot0: I'm not *against* this change, strictly speaking (happy Firefox user), but it feels a bit unrelated to test-containers.

buildSrc/src/main/groovy/geb.groovy-module.gradle
```diff
@@ -80,5 +80,7 @@ tasks.named('rat') {
         '**/MANIFEST.MF', // TODO: can we remove this?
         '**/META-INF/services/**', // TODO: can we remove this?
         '**/META-INF/groovy/**', // TODO: can we remove this?
+        '**/integration-test/resources/assets/*.txt', // exclude test artifacts
+        '**/integration-test/resources/static/*.html', // exclude test artifacts
```
jonnybot0: Maybe just add the ASL notice to the top of the files? I know it's nitpicky, but it wouldn't hurt, right?

doc/manual-snippets/real-browser/src/test/resources/GebConfig.groovy
cbmarcum: Sort of.  This is a container test and I'm using macOS arm64 and there is no native Chrome arm64 image so they fail for me.  That is the reason I changed the default container browser from Chrome to Firefox.
Let me know if you prefer it not to be in this PR.

gradle/libs.versions.toml
cbmarcum: It moved to line 72 and uses a version.ref.. It should have been removed.

Dockerfile
cbmarcum: I've but it back to original.  I'll do a separate PR with the explanation for it.  You are correct, It is not required for this PR.

gradle/libs.versions.toml
cbmarcum: I changed this to just edit the line.

doc/manual-snippets/real-browser/src/test/resources/GebConfig.groovy
cbmarcum: I reverted this. I'll put it in a separate PR to make it cleaner.

buildSrc/src/main/groovy/geb.groovy-module.gradle
cbmarcum: I removed these Rat excludes and updated the tests that check byte size.

cbmarcum:
Hi @jonnybot0 , I think I've handled everything suggested.  Please have another look.

PR #321, opened 2026-04-03
change manual-snippets / real-browser tests to use a Firefox container

cbmarcum:
I'm using macOS arm64 and there is no native Chrome arm64 image so they fail for me running locally. 
There are Firefox browser containers for both amd64 and arm64.
This would also match what we're doing in geb-testcontainers when it is merged.


cbmarcum:
wrong branch

PR #322, opened 2026-04-04
change the manual-snippets / real-browser test configuration to use a…

cbmarcum:
… Firefox container.

I'm using macOS arm64 and there is no native Selenium Chrome arm64 image so they fail for me when running locally.
This configuration would match the Geb-Testcontainers tests as well.


jonnybot0:
Hey, @cbmarcum - it looks to me like the architectures available have updated. There should be support now based on the page you linked. Let me know if there's some additional blocker that we can fix by, for example. updating Selenium.

